PR #64, opened 2015-02-04
Bug 1129250 - syslog should log to papertrail using tcp/tls

gregarndt:
* Updated packages and base image
* update app image to use new base, tcp/tls settings for papertrail


gregarndt:
If approved, I can update the other worker types (will do it in the morning as to not break things during the night)

lightsofapollo:
@gregarndt lgtm but I will leave it to you to land/deply

PR #65, opened 2015-02-10
Bug 1131747 - Add testdroid proxy feature to docker-worker

gregarndt:


lightsofapollo:
@gregarndt wow I am surprised this is fairly small +1

PR #66, opened 2015-03-02
Spot node termination/new Worker-Queue interaction/cancellation

gregarndt:
This PR is a combination of multiple bugs that are open for the worker that all kind of fit into each other based on the new queue rewrite:
Bug 1080278 - docker-worker: Listen for task-exception messages for given workerGroup and workerId 
Bug 1110560 -  docker-worker: Report exception when spot node shutdown (clean shutdown) 
Bug 1126648 - docker-worker: Migrate workers to the new task polling workflow

Also, at the same time some migration work has been made for 6to5/babel.  There are still pieces to convert and variable scoping changes that can be made, but this was the beginning work and if it proves to work well in production we can convert the rest with ease.

There were minor changes made to the tests to green them up because of how the queue now reports resolved states of tasks.  That accounts for 2/3rds of the files changes.

Note: there were changes made and a file included for api reference so that the queue rewrite could be used locally while testing.  These changes will not make it into the final merge once the queue is deployed.


bin/worker.js
```diff
@@ -119,7 +120,13 @@ co(function *() {
   // Initialize the classes and objects with core functionality used by higher
   // level docker-worker components.
   config.docker = require('../lib/docker')();
-  config.queue = new taskcluster.Queue({ credentials: config.taskcluster });
+  var jsonFromUrl = JSON.parse(fs.readFileSync('test/integration/cancelTaskReference.json'));
```
gregarndt: This will be removed once the queue is deployed.  Only for local testing with the queue rewrite.

config/test.js
```diff
@@ -3,6 +3,9 @@ var path = require('path');
 module.exports = {
   testMode: true,
   createQueue: false,
+  queue: {
```
gregarndt: This will be removed once the queue is deployed.  Only for local testing with the queue rewrite.

test/integration/cancel_task_test.js
```diff
@@ -0,0 +1,37 @@
...
+import co from 'co';
+import taskcluster from 'taskcluster-client';
+import DockerWorker from '../dockerworker';
+import TestWorker from '../testworker';
+
+var jsonFromUrl = JSON.parse(fs.readFileSync('test/integration/cancelTaskReference.json'));
+
+suite('Cancel Task', () => {
+  test("cancel", async () => {
+    var CancelQueue = taskcluster.createClient(jsonFromUrl);
```
gregarndt: This will be removed once the queue is deployed.  Only for local testing with the queue rewrite.

test/integration/cancelTaskReference.json
```diff
@@ -0,0 +1,350 @@
+{
```
gregarndt: This file will be deleted before merge once the queue is deployed.

lib/features/artifacts.js
```diff
@@ -14,13 +14,12 @@ var debug = require('debug')('docker-worker:middleware:artifact_extractor');
 var format = require('util').format;
 var Promise = require('promise');
 
-function Artifacts() {}
+export default class Artifacts {
+  constructor() {}
```
lightsofapollo: do you need this?

lib/features/bulk_log.js
```diff
@@ -58,9 +59,14 @@ BulkLog.prototype = {
       }
     );
 
-     var stat = yield fs.stat.bind(fs, this.file.path);
+    let stat = await new Promise((accept, reject) => {
```
lightsofapollo: we should use mz/fs instead of wrapping these calls ourselves.

lightsofapollo: see mz package on npm

lib/features/extend_task_graph.js
```diff
@@ -28,8 +27,14 @@ ExtendTaskGraph.prototype = {
     var scheduler = taskHandler.runtime.scheduler;
 
     // Raw tar stream for the content.
+    var contentStream;
     try {
-      var contentStream = yield container.copy({ Resource: graphPath });
+      contentStream = await (new Promise((accept, reject) => {
+        return container.copy({ Resource: graphPath }, function(err, data) {
```
lightsofapollo: why is this now wrapped in a promise ?

lib/features/extend_task_graph.js
```diff
@@ -114,13 +122,13 @@ ExtendTaskGraph.prototype = {
     }
 
     // Iterate through the graphs extending where possible.
-    var extensions = payload.graphs.map(function(graph) {
-      return this.extendTaskGraph(taskHandler, graph);
-    }, this);
+    await Promise.all(payload.graphs.map((graph) => {
```
lightsofapollo: try this instead (it works)

```js
await Promise.all(payload.graphs.map(async (graph) => {
  await this.extendTaskGraph(taskHandler, graph)
}))
```

lightsofapollo: async returns a promise so your map will now return an array of promises which `Promise.all` can deal with.

lib/features/local_live_log.js
```diff
@@ -23,27 +23,27 @@ var debug = require('debug')(
 );
 
 // Alias used to link the proxy.
-function TaskclusterLogs() {
-  this.bulkLog = new BulkLog(BACKING_ARTIFACT_NAME);
-}
+export default class TaskclusterLogs {
+  constructor() {
+    this.container = null;
```
lightsofapollo: This statement does not really do anything but might be here for doc purposes?

test/testworker.js
```diff
@@ -28,57 +28,63 @@ var schedulerEvents = new (require('taskcluster-client').SchedulerEvents);
...
+    // and container names must start with an alphanumeric character.
+    this.workerId = workerId || this.workerType.replace(/^[_-]*/, '');
+    this.worker = new Worker(PROVISIONER_ID, this.workerType, this.workerId);
+
+    this.pulse = config.get('pulse');
 
-TestWorker.prototype = {
-  __proto__: EventEmitter.prototype,
+    this.queue = new Queue({
+      baseUrl: 'http://localhost:60001/v1',
```
gregarndt: this will be removed when queue is deployed

lib/host/aws.js
```diff
@@ -92,7 +94,13 @@ function* configure (baseUrl) {
   return config;
 };
 
+function* getSpotTerminationTime() {
```
lightsofapollo: naming nit: We have generic names in this file for getTerminationTime is probably better then including spot in the name here.

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
+import xml2js from 'xml2js';
```
lightsofapollo: @jonasfj Please take closer look here just doing some skimming

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+  async getTasksFromQueue(queue, numberOfTasks) {
+    let tasks = [];
+    let uri = `${queue.signedPollUrl}&numofmessages=${numberOfTasks}`;
+    debug(`requesting: ${uri}`);
+    let response = await request
+      .get(uri)
+      .buffer()
+      .end();
+
+    let xml = await new Promise((accept, reject) => {
```
lightsofapollo: this is not terrible but I am tempted to say use denodeify and make parseString a const instead 

```
const parseXmlString = denodeify(xml2js.parseString.bind(xml2js))
```

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+      xml2js.parseString(response.text, (err, json) => {
+        err ? reject(err) : accept(json)
+      });
+    });
+
+    if(!xml.QueueMessagesList) { return tasks }
```
lightsofapollo: nit: either do this on multiple lines or kill the {}

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+
+    for(let message of xml.QueueMessagesList.QueueMessage) {
+      let payload = new Buffer(message.MessageText[0], 'base64').toString();
+      let task = JSON.parse(payload);
+      task.deleteUri = queue.signedDeleteUrl
```
lightsofapollo: Can you add some comments here looking at this hurts my eyes a bit (though I think I understand what is going on here)

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+       .replace('{{messageId}}', encodeURIComponent(message.MessageId[0]))
+       .replace('{{popReceipt}}', encodeURIComponent(message.PopReceipt[0]));
+      // If the task has been dequeued a lot, chances are the task is bad and should
+      // be removed from the queue.
+      let dequeueCount = parseInt(message.DequeueCount[0]);
+      if (dequeueCount >=15) {
```
lightsofapollo: This should be configurable.

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+        await this.deleteTaskFromQueue(task);
+        continue;
+      }
+
+      tasks.push(task);
+    }
+    return tasks;
+  }
+
+  async deleteTaskFromQueue(task) {
```
lightsofapollo: The control flow here is a bit awkward perhaps:

```js
let response;
try {
  response = await request.del(task.deleteUri).end()
} catch (e) {
  throw new Error(`Could not delete task from queue raw error \n ${e.stack}`)
}

if (!response.error) throw new Error(response.error.message)
```

lib/shutdown_manager.js
```diff
@@ -91,7 +101,20 @@ ShutdownManager.prototype = {
...
     }
-  }
+  },
+
+  spotNodeTerminated: co(function* () {
+    clearTimeout(this.spotTerminationTimeout);
+    var terminated = yield this.host.getSpotTerminationTime();
+
+    if (terminated) {
+      this.emit('shutdownManager:nodeTermination', terminated);
```
lightsofapollo: why prefix this if the event is on the object itself?

lib/states.js
```diff
@@ -28,24 +30,33 @@ Each "state" in the lifetime is a well defined function:
...
-    return yield this.hooks.
-      filter(hasMethod.bind(this, method)).
-      map(function(hook) {
-        return hook[method](task);
-      });
-  },
+  async _invoke(method, task) {
+    let hooks = this.hooks.filter(hasMethod.bind(this, method));
+    let links = await Promise.all(hooks.map((hook) => {
+        return new Promise(async (accept, reject) => {
```
gregarndt: Will make changes here to just do hooks.map(async (hook) => { ... instead of returning a new Promise (which asyc will do)

gregarndt: I could also just do "return await Promise.all"

lib/task.js
```diff
@@ -351,10 +435,64 @@ Task.prototype = {
...
+    this.taskState = 'aborted';
+    this.taskException = reason;
+    this.runtime.stats.increment('tasks.aborted');
+    this.runtime.log('abort task', {
+      taskId: this.status.taskId,
+      runId: this.runId,
+      reason: reason
+    });
+
+    if (this.dockerProcess) { this.dockerProcess.kill(); }
```
gregarndt: will remove unnecessary {}

lib/task.js
```diff
@@ -351,10 +435,64 @@ Task.prototype = {
...
+  cancel(reason) {
+    this.taskState = 'canceled';
+    this.taskException = reason;
+
+    this.runtime.stats.increment('tasks.canceled');
+    this.runtime.log('cancel task', {
+      taskId: this.status.taskId, runId: this.runId
+    });
+
+    if (this.dockerProcess) { this.dockerProcess.kill(); }
```
gregarndt: will remove {}

lib/task.js
```diff
@@ -1,33 +1,38 @@
...
+let debug = new Debug('runTask');
+let wordwrap = new Wordwrap(0, 80, { hard: true });
+
+let PAYLOAD_SCHEMA =
   'http://schemas.taskcluster.net/docker-worker/v1/payload.json#';
 
 // This string was super long but I wanted to say all these thing so I broke it
 // out into a constant even though most errors are closer to their code...
-var IMAGE_ERROR = 'Pulling docker image "%s" has failed this may indicate an ' +
+let IMAGE_ERROR = 'Pulling docker image "%s" has failed this may indicate an ' +
```
lightsofapollo: this could also be const but let is fine

lib/task.js
```diff
@@ -189,47 +198,108 @@ Task.prototype = {
...
+   * Resolves a run and reports the completion to the proper exchange.
+   *
+   * If a task was aborted because of a malformed payload or worker shutdown,
+   * the task will be reported as an exception.
+   *
+   * No Action will be taken for a canceled task because it has already been canceled
+   * by an external entity. Canceling the task run within the worker allows resources
+   * to be cleaned up.
+   */
+  async completeRun(success) {
```
lightsofapollo: Hrm there is a lot of if/else complexity here can we break this up into more functions and/or early returns?

lib/task_listener.js
```diff
@@ -13,166 +13,254 @@ var request = require('superagent-promise');
...
+    super();
+  }
 
-TaskListener.prototype = {
-  __proto__: EventEmitter.prototype,
+  listenForShutdowns() {
+    // If node will be shutdown, stop consuming events.
+    if (this.runtime.shutdownManager) {
+      this.runtime.shutdownManager.once(
+        'shutdownManager:nodeTermination', () => {
```
lightsofapollo: see note above about unusual name here.

lib/task_listener.js
```diff
@@ -13,166 +13,254 @@ var request = require('superagent-promise');
...
+    if (reason !== 'canceled') return;
+
+    var taskId = message.payload.status.taskId;
+    var task = this.runningTasks.find(
+      (task) => { return (task.status.taskId === taskId && task.runId === runId); }
+    );
 
-  connect: function* () {
-    this.runtime.log('listener connect');
+    if (!task) { debug('task not found to cancel'); return;}
```
lightsofapollo: nit: get rid of {} or multiline

lib/task_listener.js
```diff
@@ -13,166 +13,254 @@ var request = require('superagent-promise');
...
+          return await this.claimTask(task);
+        }));
+        newClaims.forEach(this.runTask.bind(this));
+        claims = claims.concat(newClaims);
+      }
+    }
+    debug(`Claimed ${claims.length} tasks`);
+  }
+
+  scheduleTaskPoll(nextPoll=5000) {
```
lightsofapollo: polling interval should be configurable

test/integration/alivenss_check_test.js
```diff
@@ -7,7 +7,10 @@ suite('Aliveness check', function() {
   var TestWorker = require('../testworker');
 
   // Ensure we don't leave behind our test configurations.
-  teardown(settings.cleanup);
+  teardown(co(function* () {
```
lightsofapollo: still need co?

test/integration/cancel_task_test.js
```diff
@@ -0,0 +1,37 @@
...
+        features: {
+          localLiveLog: false
+        },
+        maxRunTime: 60 * 60
+      }
+    };
+    var taskId = slugid.v4();
+    var worker = new TestWorker(DockerWorker);
+    let canceledTask;
+    worker.on('task run', co(function* () { yield queue.cancelTask(taskId); }));
```
lightsofapollo: nit: new use of co?

lib/task_listener.js
```diff
@@ -13,166 +13,254 @@ var request = require('superagent-promise');
...
+        this.runtime.queue.claimTask(task.taskId, task.runId, this.claimConfig)
+      );
+      await this.taskQueue.deleteTaskFromQueue(task);
+    }
+    catch (e) {
+      // Server error or 401 Authentication errors should stop trying to claim tasks
+      // and not delete the message from the queue
+      if (e && ((500 <= e.statusCode && e.statusCode < 600) || e.statusCode === 401)) {
+        throw e;
       }
```
jonasfj: If you don't throw an error here... you should explicitely `return undefined` or `null`... Clearly you don't have a claim.

You might increment a counter for this... I don't know...

lib/task_listener.js
```diff
@@ -13,166 +13,254 @@ var request = require('superagent-promise');
...
+      // Move onto the next queue if more tasks are needed but current queue is exhausted
+      // This ensures that as many tasks as possible are consumed from the highest priority
+      // queue
+      while(claims.length < availabileCapacity) {
+        let tasksNeeded = availabileCapacity - claims.length;
+        let tasks = await taskQueue.getTasksFromQueue(queue, availabileCapacity-claims.length);
+        if (!tasks.length) break;
+        let newClaims = await Promise.all(tasks.map(async (task) => {
+          return await this.claimTask(task);
+        }));
```
jonasfj: Not all `queue.claimTask` are successful... Some returns a 4xx error that should be ignored. Please do filter out null properties from the `newClaims` array... Refer to my comment on `this.claimTask` about explicitely returning null.

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+      xml2js.parseString(response.text, (err, json) => {
+        err ? reject(err) : accept(json)
+      });
+    });
+
+    if(!xml.QueueMessagesList) { return tasks }
+
+    for(let message of xml.QueueMessagesList.QueueMessage) {
+      let payload = new Buffer(message.MessageText[0], 'base64').toString();
+      let task = JSON.parse(payload);
```
jonasfj: I know that the azure queue contains messages (so that name is taken), but can please not call these for_tasks_.
Maybe advice, and wrap them in a class so we can make`deleteTaskFromQueue` into a method`..
```
Class Advice {
  constructor(payload, deleteUrl) {...}
  async remove() {try {await request.del(this.deleteUri)....} }
}
```

The azure queues contains advice or hints that a task _may_ be be pending. It might also not be pending or even exist. Hence calling it _advice_ might work well.

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+  constructor(config) {
+    this.queues;
+    this.workerType = config.workerType;
+    this.provisionerId = config.provisionerId;
+    this.client = config.queue;
+    this.log = config.log;
+  }
+
+  async getQueues() {
+    if (!this.queues || (new Date(this.queues.expires)).getTime() > Date.now()) {
```
jonasfj: How about giving yourself 5 min to pull the queue... Ie.
`new Date(this.queues.expires)).getTime() > Date.now() + 5 * 60 * 1000`

Ideally we should validate this before every request. But that non-trivial the queue may return a different number of queues next time... Anyways, 5 min should enough to poll the entire list.

lib/features/local_live_log.js
jonasfj: Makes a lot of sense... It could also be made a member using `container: null`, as `null` is a value type. However, this is a decent pattern, as setting attributes s.t. `constainer: null` only works well for value types, it is a footgun for objects..

lib/host/aws.js
```diff
@@ -20,11 +20,13 @@ AWS Metadata service endpoint.
 @const
 @see http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AESDG-chapter-instancedata.html
 */
-var BASE_URL = 'http://169.254.169.254/2012-01-12';
+var BASE_URL = 'http://169.254.169.254/latest';
```
jonasfj: Is this deliberate? I suspect you just fixed it to `latest` inorder to get the spot shutdown notifications.
Note, you can probably list available versions with `curl http://169.254.169.254/` on ec2.

We can also just hope aws maintains backwards compatibility, as they have so far. We're not going to see breakage when they add more keys to the meta-data service.

lib/queueservice.js
```diff
@@ -0,0 +1,82 @@
...
+      this.queues = await this.client.pollTaskUrls(this.provisionerId, this.workerType);
+    }
+    return this.queues.queues;
+  }
+
+  async getTasksFromQueue(queue, numberOfTasks) {
```
jonasfj: How about just passing in an integer instead of the actual queue?
Then the list of queue objects shouldn't have the option of leaving this class. Abstracting that logic away would be neat.

Also then you can do the check of signed url expiration and potential call to `pollTaskUrls` before polling the azure queue, a head of each and every poll...
Note, this way you might accept that it only has to be 3 min a head of expiration. Before refreshing, and chance of making a request with expired signature is much lower.

Note, you should handle the special case where the integer given is higher than the number of queues, by doing integer modulo number of queues. This could happen if after refreshing you get fewer azure queues than before.

Note in this setup, all you return to `TaskListener` from `getQueues` is the number of queues that exists. Granted you might want to change `getQueues` into `getNumberOfQueues`.

lib/queueservice.js
jonasfj: Also, as this will always return `[]`, just write `return [];` it more obvious that way.

lib/states.js
jonasfj: The decent two-liner is:
```js
let hooks = this.hooks.filter((hook) => { return hook[method] instanceof Function });
return Promise.all(hooks.map((hook) => { return hook[method](task); }));
```
Maybe one can do the `filter` on the same line, I couldn't make it readable :)

lib/task.js
```diff
@@ -189,47 +198,108 @@ Task.prototype = {
...
+    let queue = this.runtime.queue;
+
+    if (this.taskException) {
+      if (this.taskState !== 'canceled') {
+        await this.runtime.stats.timeGen(
```
jonasfj: Do we really want statistics for all calls to the queue?
If so, which might be valid... Can we not just make a hack and wrap the methods on the `taskcluster.Queue` instance?
So we don't have to read about it anymore...
Example:
```js
let queue = new taskcluster.Queue(...);
this.runtime.stats.wrapQueueClient(queue);
....
/** Wrap methods of a taskcluster client library */
function wrapQueueClient(client) {
  ['reportCompleted', 'reportException', ...].forEach((name) => {
    let method = client[name];
    client[name] = function(...args) {
      return timeGen('tasks.time.' + name, method.apply(this, args);
    };
  });
};
```

I'm also open to adding support for statistics collection to `taskcluster-client`. The client object to just emit an event whenever a request is started and whenever one is finished. The finished one could even include timestamp, if you would like. Anyways, abstracting this away with monkey patching is better than having it spread through the code base.

lib/task.js
jonasfj: Agree.. or at the very least use the methods you added like `isAborted` I think I saw you adding that to this object.
Also you might want to rename `completeRun` into something like `finishRun`, or anything that isn't _complete_ :)

lib/task.js
```diff
@@ -351,10 +435,64 @@ Task.prototype = {
...
+
+  /**
+   * Aborts the running of the task.  This is similar to cancelling a task, but
+   * will allow time to upload artifacts and report the run as an exception instead.
+   *
+   * @param {String} reason - Reason for aborting the test run (Example: worker-shutdown)
+   */
+  abort(reason='aborted') {
+    this.taskState = 'aborted';
+    this.taskException = reason;
```
jonasfj: Please explain the default string? It's not a valid _reason_ as far as I'm aware.
Valid exceptions that a worker can report is: `worker-shutdown` and `malformed-payload`.

lib/task.js
```diff
@@ -351,10 +435,64 @@ Task.prototype = {
...
+   * entity and has already been published to task-exception exchange.  This will
+   * kill the docker container that might be running, attempt to release resources
+   * that were linked, as well as prevent artifacts from uploading, which cannot
+   * be done after a run is resolved.
+   *
+   * @param {String} reason - Reason for cancellation
+   */
+  cancel(reason) {
+    this.taskState = 'canceled';
+    this.taskException = reason;
```
jonasfj: Again I don't see how reason makes any sense here.

If cancelled, you may not report state... Or at least you shouldn't it'll just give you an error.

jonasfj:
@gregarndt, I got through most of it at a high level... I'll probably have to take another trip through tomorrow - this is a lot of changes. I see the `taskState` isn't awesome, but i'm not sure how else to solve. Maybe I'll get a good idea next time I look at it.

Anyways, don't let my desire to take a deep look stop you from addressing comments.

lib/host/aws.js
gregarndt: Yea, this was deliberate.  When I curled for version, the most recent one was 11-05-2014 and this seems to be added after that.

lib/features/extend_task_graph.js
gregarndt: mistake in looking at some dockerode code that was callback based and forgetting we are wrapping this in a promise proxy elsewhere.  Will fix it up.

lib/queueservice.js
gregarndt: hrm, I think I had this wrong, is should be < rather than greater than.  Basically wanted to find out if the time now is beyond the expiration.  So, in your suggestion we would want to do a refresh if within 5 minutes of the expiration, right?

lib/queueservice.js
gregarndt: The thought I was having here is getting the queues and querying the queues were two very dependent, often sequential, actions and didn't rely on some internal state of that class (list of queues).  Basically I want to know what queues exist, it tells me, and then for each of those I want to do something with them. If the task listener wants to log something about the queues, it already has the information because it has a queue object rather than some integer representing some state within this class that may or may not be the same depending on integer and when the queue refresh happens.
Definitely would need some handling around higher integer given but a lower number of queues because of a refresh.  If there is the possibility of the number of queues going down on a refresh, does the possibility also exist for the number to go up?  then the tasklistener wouldn't know about those until the next call to getTasksFromQueue..

lib/task_listener.js
gregarndt: awesome catch, I meant to return undefined or something similar.  Thanks!

lib/task.js
gregarndt: Ah, this was just following what was already an existing way of collecting these stats without changing it up too much.  I definitely like the idea of either monkey patching the queue or adding something to the client.  Having timeGen and similar everywhere is very ugly.

lib/task.js
gregarndt: I wish I could think of a reason why I defaulted, this but it escapes me.

gregarndt:
Thanks for the reviews guys and look forward to any other comments you have.  I'll work through them in the morning. @jonasfj taskState is awful, I agree.  I think @lightsofapollo and me are going to have a race to see who gets sick of it the quickest and turn task.js into a FSM of some sort.

lib/features/artifacts.js
gregarndt: fixed in ccf932af0797e4aeff03774db82770f116158cd2

lib/features/bulk_log.js
gregarndt: Thanks for pointing out mz, didn't look at it before...
fixed in 56188c9248b3bb0cfc6b78a75fb5b6334af0b59f

lib/features/extend_task_graph.js
gregarndt: fixed in 56188c9248b3bb0cfc6b78a75fb5b6334af0b59f

lib/features/extend_task_graph.js
gregarndt: fixed in 56188c9248b3bb0cfc6b78a75fb5b6334af0b59f

lib/host/aws.js
gregarndt: fixed in 56188c9248b3bb0cfc6b78a75fb5b6334af0b59f

lib/shutdown_manager.js
gregarndt: fixed in 56188c9248b3bb0cfc6b78a75fb5b6334af0b59f

lib/task.js
gregarndt: fixed 22c327c5c3a0bdaeff33a29231384aa64b058415

lib/task.js
gregarndt: So I separated out the two concerns...completing a run and aborting one.  Completing a run should only be concerned with a run that hasn't actually been aborted, otherwise abortRun() should do the right thing of either reporting an exception (worker-shutdown, malformed payload) , or reporting a failure for various reasons (docker issue, etc)

Changed in 22c327c5c3a0bdaeff33a29231384aa64b058415

lib/task_listener.js
gregarndt: fixed 56188c9248b3bb0cfc6b78a75fb5b6334af0b59f

lib/queueservice.js
jonasfj: Perhaps an even better thing to do would be to move the `claimTask` part from the `task_listener.js` into the `queueservice.js`... Then this file could encapsulate all the code that deals with pulling azure messages, iterating over azure queues and claiming tasks based on messages.

The task listener also deals with other stuff like, waiting for capacity to be available, scheduling polling for tasks and listening for cancels. 

The main reason I suggested exposing integers instead of objects was to reduce the interface between the files. Encapsulation is key and javascript one has to be careful to make sure it gets done. Putting all the polling logic and claim task logic in the same place makes sense. While still have the scheduling for polling of tasks in the task_listener. (my two cents)

lib/task_listener.js
```diff
@@ -13,166 +13,275 @@ var request = require('superagent-promise');
...
+      credentials: this.runtime.pulse
     });
 
-    yield listener.bind(queueEvents.taskPending({
+    await cancelListener.bind(queueEvents.taskException({
+      workerId: this.runtime.workerId,
       workerType: this.runtime.workerType,
+      workerGroup: this.runtime.workerGroup,
       provisionerId: this.runtime.provisionerId
     }));
```
jonasfj: We shouldn't crash on failure to keep the connection to pulse open. As we're really _only_ using it for cancel messages (which is best effort), we should listen for `cancelListener.on('error', (err) => {... });`.

Ideally we could restart the connection and things, but it's a lot of work and it seems better to do this at `PulseListener` level. I want to add support for it there I just haven't figured out how to do so yet; it should probably be hidden behind a option.

Anyways, IMO we should just log the error and then ignore it.. Accepting that for the rest of this docker-worker lifecycle we're not listening for taskExceptions anymore. Then we can add the reconnect option when we implement that in `PulseListener`.

lib/task_listener.js
```diff
@@ -13,166 +13,275 @@ var request = require('superagent-promise');
...
 
-    // Share the queue between all workerTypes of the same provisioner.
-    var queueName;
-    if (this.runtime.createQueue) {
-      queueName =
-        QUEUE_PREFIX +
-        this.runtime.provisionerId + '/' + this.runtime.workerType;
-    }
+    this.runtime.log('cancelling task', {taskId: message.payload.status.taskId});
+    task.cancel(reason);
```
jonasfj: I'm not sure you are aware of `PulseListener` semantics. The event handler may return a promise.
If the promise is rejected, `PulseListener` will reject the event and the event will happen again, second time it fails it'll be logged and discarded.

So if you wanted to you could return a promise from `task.cancel()`. Though I'm not sure this is always desired. So maybe you have the right semantics :)

jonasfj:
Over all starting to look good. A few minor issues left like, let's not crash on pulse connection error (it's TCP).

I would also love to see better separation of `queueservice.js` and `task_listener.js`, ideally `queueservice.js` only returns claims or something like that. To keep logic encapsulated.
(Also makes it easier to hack if one day we change semantics on priority).

lib/queueservice.js
```diff
@@ -0,0 +1,85 @@
...
+import request from 'superagent-promise';
+import denodeify from 'denodeify';
+
+const parseXmlString = denodeify(xml2js.parseString.bind(xml2js));
+
+let debug = Debug('taskcluster-docker-worker:queueService');
+
+export default class TaskQueue {
+  constructor(config) {
+    this.queues;
```
jonasfj: So a tip I use to make sure that config errors are cached quickly is to validate the input of variables like `config`.
Clearly this is not performance critical, but it's a lot easier to find mismatches between expected input and actual input, if it's validated.

Other wise, those mismatches/bugs shows up at random places during execution which is less fun. Just a general pattern I would recommend, especially when cross module boundaries.
Random example, see doc comment and assert in [queue reaper](https://github.com/taskcluster/taskcluster-queue/blob/1b591a0d03cab96c8b1c9a6ece353e37d44563d9/queue/reaper.js#L8-L35).
Whether you use assert, should.js, expect.js, joi or JSON schemas I don't care... Personally I'm starting to lean towards should.js or expect.js (expect is worse, but should.js don't work well with `undefined`/`null`, it's just too monkey patchy..).

jonasfj:
Fix my [last comment](https://github.com/gregarndt/docker-worker/commit/d7cfb93fc49487b12419a6880a33e323682eebd7#commitcomment-10075140), both the control flow and the log message (so it's clear that this is a non-bug special condition), and this is good to go!

It looks really decent with the encapsulation now.

lib/queueservice.js
```diff
@@ -0,0 +1,231 @@
...
+  async claimTask(task) {
+    try {
+      let claim = await this.stats.timeGen(
+        'tasks.time.claim',
+        this.queue.claimTask(task.taskId, task.runId, this.claimConfig)
+      );
+
+      this.stats.increment('tasks.claims');
+
+      await this.deleteTaskFromQueue(task);
```
jonasfj: Actually we can just move this down past the `catch`-block.

PR #69, opened 2015-03-13
Bug 1142697 - Containers should only be removed after an expiration

gregarndt:


lib/volume_cache.js
```diff
@@ -103,14 +103,27 @@ export default class VolumeCache {
...
           var cacheKey = cacheName + KEY_DELIMITER + instance;
           var instancePath = this.cache[cacheName][instance].path;
           var statName = 'cache.volume.' + cacheName + '.instance_removed';
-          await this.stats.timeGen(statName, removeDir(instancePath));
+          // Remove instance from the list of managed caches so another worker
+          // does not try to claim it.
           delete this.cache[cacheName][instance];
-          this.log('cache volume removed',
-            {key: cacheKey, path: instancePath});
+          try {
```
lightsofapollo: nit: this block is already somewhat complicated I would extract this logic into another method (maybe tryCacheRemoval())

PR #73, opened 2015-03-27
Isolated mode

lightsofapollo:


test/integration/isolated_mode_test.js
```diff
@@ -0,0 +1,46 @@
+suite('Shutdown on idle', function() {
+  var co = require('co');
```
gregarndt: for new tests that we write, do we want to still use co or follow babel/es6 conventions?

deploy/bin/configure.js
```diff
@@ -42,10 +42,6 @@ function* configure() {
     currentConfig[key] = (yield prompt(humanDesc)) || defaultValue;
   }
 
-  console.log();
-  console.log(util.inspect(currentConfig, { colors: true }));
-  console.log();
-
   // Yeah bad things will happen if rejected too often...
   if (!(yield prompt.confirm("Does this look right? "))) {
```
gregarndt: We removed outputting the contents of the config to confirm if it looks correct, do we want to remove this prompt as well?

lightsofapollo-taskcluster: ah!! this is a mistake now that I think about it.

test/integration/isolated_mode_test.js
lightsofapollo-taskcluster: your right i am being lazy here :) bad copy/paste!

PR #75, opened 2015-03-30
Bug 1149066 max tasks

gregarndt:


config/defaults.js
```diff
@@ -69,7 +69,10 @@ module.exports = {
     // Number of times to retry requests to the task queue
     maxRetries: 5,
     // Amount of time to wait between retries
-    requestRetryInterval: 2 * 1000
+    requestRetryInterval: 2 * 1000,
+    // Azure currently has a max of 32 messages that can be retrieved from the
+    // queue at a time.
+    maxMessagesPerRequest: 32
```
jonasfj: It's a detail, but this is more of a constant than a configurable property..

I would hard code it into `lib/queueservice.js` as a constant with something like:
```js
/** Maximum number of message that can retried from azure per request */
const MAX_MESSAGES_PER_REQUEST = 32;
```

Or something like that...

gregarndt: Funny you mention that, I originally had it as a const but then was unsure if this should be configurable.  Thought that perhaps there might be a time where we might want to decrease/increase this, but was unsure.

PR #78, opened 2015-04-16
Bug 1056384 - Maxruntime adjustment

gregarndt:
Max runtime should include killing/stopping states

If max run time is reached while stopping states (specifically artifact copy/upload), the state handler will abort and the task can clean up and resolve as an exception.

### DIFF
--- a/lib/features/artifacts.js
+++ b/lib/features/artifacts.js
@@ -71,7 +71,18 @@ export default class Artifacts {
 
       return;
     }
+
+    if (taskHandler.isAborted()) return;
     var tarExtract = tarStream.extract();
+
+    // If abort event is emitted, shut down the tarExtract stream which will cause
+    // the file to abort uploading and return back to the state handler.
+    taskHandler.on('abort', () => {
+      debug('Task has been aborted. Destroying tar stream.');
+      tarExtract.destroy();
+      tarExtract.emit('finish');
+    });
+
 
     // Begin unpacking the tar.
     contentStream.pipe(tarExtract);
@@ -142,7 +153,9 @@ export default class Artifacts {
       var res = yield waitForEvent(putReq, 'response');
 
       // If there was an error uploading the artifact note that in the result.
-      if (res.error) {
+      // Assume that if the task has been aborted, the upload will have failed
+      // and an error is expected
+      if (res.error && !this.taskHandler.isAborted()) {
         taskHandler.stream.write(taskHandler.fmtLog(
           'Artifact "%s" failed to upload "%s" error code: %s',
           name,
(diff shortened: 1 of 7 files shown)


lib/features/artifacts.js
```diff
@@ -71,8 +71,19 @@ export default class Artifacts {
 
       return;
     }
+
+    if (taskHandler.isAborted()) return;
```
jonasfj: you might want to try `contentStream.destroy()` or something... not sure the right thing is...

lib/features/artifacts.js
```diff
@@ -142,7 +153,9 @@ export default class Artifacts {
       var res = yield waitForEvent(putReq, 'response');
```
jonasfj: This is scary... what if a repsonse never arrives? Then we hang here forever...

jonasfj: Am I right?

Also can we move away from `co` :)

jonasfj: hmm... Maybe we need to refactor all of this a bit... Into something with abstractions... maybe:
`getFileStreamFromContainerFile(filePath, async (length, stream) => { /* upload logic */ }, , retries = 5)` and for directories:
`getFileStreamsFromContainerFolder(folderPath, async (length, stream) => { /* upload logic (called once for each task) */ }, retries = 5)`.

So that `getFileStreamFromContainerFile` handles get file... and tarextract etc... plus retries if needed... In folder it has to track which files it already uploaded successfully and just do `.drain()` for those on next try... well maybe.. I can see the best abstractions right now...

gregarndt: As far as co() it will be gone soon.  We need to rewrite the artifact handler and I was doing this in stages.

gregarndt: Well, in the case of a regular upload, you're right, it could hang forever.  In the case of an abort, the uploadArtifact method will return and the task will complete, however there will be lingering things outside of the task waiting for that response still.

jonasfj: using abort timeout is nice... but uploads shouldn't hang they need to be retried if they timeout (< 1 mb/s for 60s).
Or if network connection is broken...

lightsofapollo-taskcluster: you probably want to race between completed and an error.

lib/task.js
```diff
@@ -532,9 +546,22 @@ export default class Task {
...
+    let links;
+    try {
+      links =
+        await stats.timeGen('tasks.time.states.linked', this.states.link(this));
+    }
+    catch (e) {
+      // If this state threw an exception because the task is being aborted,
+      // debug the error and continue where it will be aborted properly
+      if (this.isCanceled() || this.isAborted()) {
+        debug(e);
```
lightsofapollo-taskcluster: nit: ideally we could prefix this...

PR #80, opened 2015-04-17
Spot termination should be checked on startup

gregarndt:
This fix will do a few different things:
1. Only start docker-worker when docker is started and the run level is 2, 3, 4, or 5.  This will help both in the cases of spot terminations but also when the worker attempts to shut itself down.
2. Check the spot termination endpoint when starting up and if a termination event is detected, the capacity will be set to 0.


lib/host/aws.js
```diff
@@ -49,33 +51,44 @@ Read AWS metadata and user-data to build a configuration for the worker.
...
+    getText(baseUrl + '/meta-data/public-hostname'),
+    // workerId
+    getText(baseUrl + '/meta-data/instance-id'),
+    // workerType
+    getText(baseUrl + '/meta-data/ami-id'),
+    // workerGroup
+    getText(baseUrl + '/meta-data/placement/availability-zone'),
+    // workerNodeType
+    getText(baseUrl + '/meta-data/instance-type')
+  ]);
```
jonasfj: There is no real need for parallel requests here... We only do this one startup...
so you have been off fine with `await getText(baseUrl + '/meta-data/public-hostname')` for all the config entries..

Sometimes I wish we had things like`.spread` and `.props`: https://github.com/petkaantonov/bluebird/blob/master/API.md#props---promise

Anyways, this is very good :) Don't change it!

jonasfj:
Did a light review... This looks fine... I didn't look too much at the test cases... let me know if you want to a deeper review, I think you can land this so r+

PR #81, opened 2015-04-22
Small fixes

petemoore:


petemoore:
@gregarndt very small fixes here - a link has moved, and a fix so you can call build.sh without needing to cd into parent directory first.

gregarndt:
Looks good to me.

PR #82, opened 2015-04-23
Check diskspace threshold against a mimimum capacity

gregarndt:
If current capacity is maxed out, ensure that at last a buffer of threshold * 1 is available when checking diskspace treshold.


jonasfj:
r+

PR #85, opened 2015-04-27
Allow video and audio devices to be mounted within container

gregarndt:
This will allow audio and video loop back devices to be mounted within a task container using the snd-aloop and v4l2loopback kernel modules respectively.

A task payload can specify the devices necessary and if a device manager is registered for that type of device, they will be mounted within the container.


lib/task.js
```diff
@@ -99,6 +99,23 @@ async function buildVolumeBindings(taskVolumeBindings, volumeCache, taskScopes)
...
+function buildDeviceBindings(devices) {
+  let deviceBindings = [];
+  for (let deviceType in devices) {
+    let device = devices[deviceType];
+    device.mountPoints.forEach((mountPoint) => {
+      deviceBindings.push(
+        {
+          PathInContainer: mountPoint,
+          PathOnHost: mountPoint,
+          CgroupPermissions: 'rwm'
```
gregarndt: this could probably be chanced to be more restrictive, (will test out)

lib/task.js
```diff
@@ -99,6 +99,23 @@ async function buildVolumeBindings(taskVolumeBindings, volumeCache, taskScopes)
   return [caches, bindings];
 }
 
+function buildDeviceBindings(devices) {
+  let deviceBindings = [];
+  for (let deviceType in devices) {
+    let device = devices[deviceType];
+    device.mountPoints.forEach((mountPoint) => {
+      deviceBindings.push(
```
gregarndt: Shame that this couldn't be as straight forward as docker volume bindings like:
<path on host>:<path in container>:<permissions>

lib/task.js
```diff
@@ -172,9 +195,9 @@ export default class Task {
       let bindings = await buildVolumeBindings(this.task.payload.cache,
         this.runtime.volumeCache, this.task.scopes);
       this.volumeCaches = bindings[0];
-      procConfig.start.Binds = bindings[1];
+      procConfig.create.HostConfig.Binds = bindings[1];
     }
-
+    debug(procConfig);
```
gregarndt: removed

lib/task.js
```diff
@@ -494,6 +517,7 @@ export default class Task {
       await stats.timeGen('tasks.time.states.created', this.states.created(this));
     }
     catch (e) {
+      debug(e.stack);
```
gregarndt: removed

lib/task.js
```diff
@@ -560,6 +584,7 @@ export default class Task {
       let error = this.fmtLog('Docker configuration could not be ' +
         'created.  This may indicate an authentication error when validating ' +
         'scopes necessary for using caches. \n Error %s', e);
+      debug(e.stack);
```
gregarndt: gone

lib/task.js
```diff
@@ -99,6 +99,23 @@ async function buildVolumeBindings(taskVolumeBindings, volumeCache, taskScopes)
...
 }
 
+function buildDeviceBindings(devices) {
+  let deviceBindings = [];
+  for (let deviceType in devices) {
+    let device = devices[deviceType];
+    device.mountPoints.forEach((mountPoint) => {
+      deviceBindings.push(
+        {
+          PathInContainer: mountPoint,
```
gregarndt: Thought... specify the path that it should be mounted to within the container like we do for volume mounts.

lightsofapollo-taskcluster: ahh- we could attempt to mount to default 1-2/etc.. locations but it appears that our test harnesses are smart enough to pick any number

lib/task_listener.js
```diff
@@ -17,18 +17,29 @@ var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 var TaskQueue = require('./queueservice');
 var exceedsDiskspaceThreshold = require('./util/capacity').exceedsDiskspaceThreshold;
+var VideoDeviceManager = require('./video_device_manager');
+var AudioDeviceManager = require('./audio_device_manager');
+
+const DEVICE_MANAGERS = {
+  'loopbackVideo': new VideoDeviceManager(),
+  'loopbackAudio': new AudioDeviceManager()
```
lightsofapollo-taskcluster: this is nicer then the CPU stuff I did you could probably refactor that out in a similar way

test/integration/parallel_worker_test.js
```diff
@@ -15,12 +15,11 @@ suite('Parallel workers', function() {
...
 
   var workerA, workerB;
-  var workerType = `test_worker_${slugid.v4()}`.substring(0,22);
 
   setup(co(function * () {
     // Each worker should use the same worker type but a unique worker id.
-    workerA = new TestWorker(DockerWorker, workerType, slugid.v4());
-    workerB = new TestWorker(DockerWorker, workerType, slugid.v4());
+    workerA = new TestWorker(DockerWorker, 'dummy-parallel-type');
+    workerB = new TestWorker(DockerWorker, 'dummy-parallel-type');
```
lightsofapollo: By making these names not unique how confident are you that we can still run these tests in parallel (i.e. your computer/my computer) or ideally eventually parallel by default on a single machine

gregarndt: Ah yea, this needs to be fixed.  This was done after Jonas and I talked about having wokertype static for the tests because of azure queues, but then I realized it was a mistake and reverted the changes I made in testworker.js but not here.  thanks for the catch.

PR #86, opened 2015-05-01
Bug 1160320 - Run containers in privileged mode

gregarndt:
This allows docker-worker to start containers in a privileged mode to allow them privileges such as running docker-in-docker which helps for running the docker worker tests.  

Both the worker and the task need to be configured to run as privileged for this to happen and tasks will be failed if they request privileged mode but the worker is not configured to allow it.

This starts the work of adding a "capabilities" object to the task payload that will tasks to define certain options that need to be available/enabled for the container.  Later this will be expanded to include things for loopback devices, possible memory limits, etc.


lib/task.js
```diff
@@ -140,8 +140,19 @@ export default class Task {
         this.task.payload, this.status.taskId
     );
 
+    let taskCapabilities = this.task.payload.capabilities || {};
+    let privilegedTask = taskCapabilities.privileged || false;
+    if (privilegedTask && !this.runtime.dockerConfig.allowPrivileged) {
```
jonasfj: I think this is solid... An alternative and/or additional way of restricting this is to require a scope for privileged tasks.
Ie. require `docker-worker:privileged-container` to be covered by `task.scopes` (they same thing we do for cache folders).

The scope could also be: `docker-worker:privileged-container:<provisionerId>/<workerType>`, since that would make it easier to issue tighter controlled scopes.

Anyways, I like that it can be completely disabled for a given workerType :)

schemas/payload.js
```diff
@@ -57,6 +57,11 @@ module.exports = {
       "description": "Caches are mounted within the docker container at the mount point specified. Example: ```{ \"CACHE NAME\": \"/mount/path/in/container\" }```",
       "type": "object",
     },
+    "capabilities": {
+      "title": "Capabilities that must be available/enabled for the task container.",
+      "description": "Set of capabilities that must be enabled or made available to the task container Example: ```{ \"capabilities\": { \"privileged\": true }```",
+      "type": "object",
+    },
```
jonasfj: Can't we type this some more... Ie. specify all properties, so it's documented what properties are allowed.
You don't have to make them required... ie:
```js
properties: {
  privileged: {
    title: "Privileged Container",
    description: "Run task in a privileged container, equivalent to running docker with `--privileged`, this only works for workerTypes configured to allow this.",
    type: "boolean",
    default: false
  }
},
additionalProperties: false,
default: {} // Specify default value, this very nice for docs, we can also use JSON schema validator
            // that substituted default value in. This is awesome because we don't have to handle the cases
            // in our code... And we don't have to hardcode `false` as default anywhere else.
            // In this case we're unlikely to ever change the default, but might change it for say cpu-shares
            // if we add such a capability later on.
```
If you decide to additionally require a scope this would also be a good place to document it.

Note, I specify `additionalProperties: false`, but I don't not specify `required: ["privileged"]`, because people don't have to specify this property.

test/docker-worker-test
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env babel-node
+
+import program from 'commander';
+import { spawn } from 'child_process';
+import path from 'path';
+import diveSync from 'diveSync';
+
+let baseDir = path.resolve(__dirname, '..');
+let binary = `${baseDir}/node_modules/mocha/bin/mocha`;
```
jonasfj: nit: I think you'll want to use `${baseDir}/node_modules/.bin/mocha`, at least that is the npm pattern... Mocha could (if they were crazy) decide to move the binary location within their package.

jonasfj: Note, if you added this script as `npm test` then `node_modules/.bin/` would be in `$PATH`. But I suspect the runner is different when running locally maybe...

Afaik it's still possible to pass arguments to `npm test`, it's just `npm test -- --this-chunk 5` but I could be remembering that wrong too...

Anyways, nit... this is not important at all...

jonasfj:
I was very lightly over the test stuff and git PR related scripts... But code changes seems sane..
My comments are optional... Maybe do that JSON schema thing, as it's very easy...

The scopes thing is a possible neat thing to do... But it's not required...

lib/task.js
jonasfj: Looking at existing docker-worker scopes... Maybe `docker-worker:capability:privileged` would be better... Adding `:<provisionerId>/<workerType>` might be nice too..

Anyways, I could imagine adding more `docker-worker:capability:...` style scopes in the future.
Example: `docker-worker:capability:cpu-shares` to allow specifying CPU-shares higher than some allowed constant....

Note, scopes are bad for things continuous resources like say cpu-shares, or memory-limit, but can still be used.. We just hard code some scope covered intervals... like `docker-worker:capability:memory-1g` and `docker-worker:capability:memory-2g`, `docker-worker:capability:memory-10g` , `docker-worker:capability:unlimited-memory`, then when someone asks for "2g" memory we have to check for one of 3 scopes...

Anyways, that's how I plan to solve it for scoping task expiration and artifact sizes when someday we do that...

PR #88, opened 2015-05-18
Https live logs

gregarndt:


config/defaults.js
```diff
@@ -122,5 +132,8 @@ module.exports = {
     password: process.env.TESTDROID_PASSWORD
   },
 
-  dockerWorkerPrivateKey: '/etc/docker-worker-priv.pem'
+  dockerWorkerPrivateKey: '/etc/docker-worker-priv.pem',
+
+  sslCertificate: '/etc/star_taskcluster-worker_net.crt',
+  sslKey: '/etc/star_taskcluster-worker_net.key'
```
jonasfj: Nit, but maybe move this under `logging`.

Or under `ssl.crtFile` and `ssl.keyFile`... I guess we'll use it for more than logging long term anyways.

gregarndt: yea, I kept it separate specifically because it could probably be used later on.  I can move it to be 
```
ssl: {
  crtFile...
}

File: lib/util/hostname.js [hunk 2]
```diff
@@ -0,0 +1,26 @@
+import assert from 'assert';
+import statelessDNSServer from 'stateless-dns-server';
+
+export default function getHostname(config, expires) {
+  let statelessConfig = config.statelessHostname || {};
+  if (!statelessConfig || !statelessConfig.enabled) {
+    return config.host;
+  }
+
+  let secret = statelessConfig.secret || '1234';
```
gregarndt: Um, this needs to be removed...whoops

test/integration/logging/local_live_log_test.js
```diff
@@ -0,0 +1,82 @@
+import co from 'co';
```
gregarndt: This will be removed.

lib/features/local_live_log.js
```diff
@@ -28,6 +29,7 @@ export default class TaskclusterLogs {
     Docker container used in the linking process.
     */
     this.container = null;
+    this.token = crypto.randomBytes(64).toString('hex');
```
jonasfj: So I thought this was a potentially bad use of entropy... Fearing that this could block for a long time, if we were very unlucky. Honestly, I would have used `slugid.v4()`, but I agree that this is not a good idea.

Anyways, looking at `uuid` underlying `slugid` I realize that it uses `crypto.randomBytes()` too...

So I dugg down to see if it was safe to pull a lot of data out of this... and it seems that it is: http://stackoverflow.com/a/18131768/68333
Basically it's the same as pseudo random, except it'll throw an error if there isn't enough entropy, but once entropy is acquired it'll never complain.

So I'm happy with this. Just wanted to add the background information here.

---
Also considering that a slugid is just as safe, why don't we just use that... It's sorter an already URL-safe base64 encoded. 16 or 64 bytes probably doesn't matter, 64 is a bit overkill; it's like we're not giving the NSA any chance at guessing right :)

lib/util/hostname.js
```diff
@@ -0,0 +1,26 @@
...
+  let domain = statelessConfig.domain || '';
+  assert(secret, 'Must supply a secret for stateless dns server');
+  assert(domain, 'Must supply a domain name used for stateless hostname');
+
+  let hostname;
+  let ip = config.publicIp || '127.0.0.1';
```
jonasfj: Shouldn't this `|| '127.0.0.1'` live in the place where publicIp is set... This seems a bit sketchy...

lib/util/hostname.js
```diff
@@ -0,0 +1,26 @@
...
+  ip = ip.split('.').map((octet) => { return parseInt(octet) });
```
jonasfj: Okay, so just because we were discussing how to do this best, I since found that, this works too:
```js
ip = ip.split('.').map(digit => parseInt(digit));
```
He he, just as an FYI... I don't care how do it here...

test/aws_host_config_test.js
```diff
@@ -34,7 +34,8 @@ suite('configuration/aws', function() {
       workerType: 'ami-333333',
       workerNodeType: 'c3.xlarge',
       workerGroup: 'us-west-2',
-      capacity: 1
+      capacity: 1,
+      publicIp: '22.33.44.252'
```
jonasfj: I have to ask? is there anything special with that IP... if so can we add a comment.

jonasfj: Ahh, maybe it's hardcoded other places in test code? if so a comment about that would be neat...

jonasfj:
Minor nits... address them or not... this looks good to go...

test/aws_host_config_test.js
gregarndt: yea, this is hardcoded in the aws_metadata which sort of acts like a mock for the meta-data endpoint.  It could be anything, just as long as it matched what's returned.

PR #89, opened 2015-05-19
updated to use docker hub registry

gregarndt:
Updates docker feature images to use docker hub instead of quay.  Also updated to use default registry for images given in the form <user>:<image>:,tag>


jonasfj:
Looks awesome to me... I'll review the other PR immediately, so hold of on deploying... for at least 30min :)

PR #90, opened 2015-05-20
Create ways of reliably inserting and removing running tasks

gregarndt:
CAnnot rely on the index position of a task because other tasks could be sliced before it adjusting the index position.  Look at task id and run ID to find the running task before cleaning up.

### DIFF
--- a/lib/shutdown_manager.js
+++ b/lib/shutdown_manager.js
@@ -89,9 +89,7 @@ export default class ShutdownManager extends EventEmitter {
     this.taskListener.on('working', this.onWorking);
 
     // Kick off the idle timer if we started in an idle state.
-    if (taskListener.pending === 0) {
-      this.onIdle();
-    }
+    if (taskListener.isIdle()) this.onIdle();
   }
 
   scheduleTerminationPoll() {
--- a/lib/task_listener.js
+++ b/lib/task_listener.js
@@ -212,23 +212,56 @@ export default class TaskListener extends EventEmitter {
     }
   }
 
+  addRunningTask(runningState) {
+    this.runningTasks.push(runningState);
+    this.incrementPending();
+  }
+
+  removeRunningTask(runningState) {
+    let taskIndex = this.runningTasks.findIndex((runningTask) => {
+      return (runningTask.taskId === runningState.taskId) &&
+        (runningTask.runId === runningState.runId)
+    });
+
+    if (taskIndex === -1) {
+      this.runtime.log('[warning] running task removal error', {
+        taskId: runningState.taskId,
+        runId: runningState.runId,
+        err: 'Could not find the task Id in the list of running tasks'
+      });
+      return;
+    }
+
+    this.cleanupRunningState(runningState);
+    this.runningTasks.splice(taskIndex, 1);
+    this.decrementPending();
+  }
+
   /**
   * Run task that has been claimed.
   */
   async runTask(claim) {
     try {
-      this.runtime.log('run task', { taskId: claim.status.taskId, runId: claim.runId });
-      this.incrementPending();
+      // Reference to state of this request...
+      let runningState = {
+        devices: {},
+        taskId: claim.status.taskId,
+        runId: claim.runId
+      };
+
+      this.runtime.log(
+        'run task',
+        {
+          taskId: runningState.taskId,
+          runId: runningState.runId
+        }
+      );
+
       // Fetch full task definition.
-      var task = await this.runtime.queue.task(claim.status.taskId);
+      var task = await this.runtime.queue.task(runningState.taskId);
 
       // Date when the task was created.
       var created = new Date(task.created);
-
-      // Reference to state of this request...
-      let runningState = {
-        devices: {}
-      };
 
       // Only record this value for first run!
       if (!claim.status.runs.length) {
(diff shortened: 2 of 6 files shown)


lib/task_listener.js
```diff
@@ -212,24 +212,57 @@ export default class TaskListener extends EventEmitter {
...
+  addRunningTask(runningState) {
+    this.runningTasks.push(runningState);
+    this.incrementPending();
+  }
+
+  removeRunningTask(runningState) {
+    let taskIndex = this.runningTasks.findIndex((runningTask) => {
+      return (runningTask.taskId === runningState.taskId) &&
+        (runningTask.runId === runningState.runId)
+    });
```
jonasfj: Curious why not just object comparison?

gregarndt: Hrm, I would like to say it was because of some great reason, but really it's just because I didn't think about it. however now thinking about it, there might be a time where something other than someone that holds the reference to runningState wants to remove the task (maybe? no idea).

lib/task_listener.js
```diff
@@ -212,24 +212,57 @@ export default class TaskListener extends EventEmitter {
...
+    if (taskIndex === -1) {
+      this.runtime.log('[warning] running task removal error', {
+        taskId: runningState.taskId,
+        runId: runningState.runId,
+        err: 'Could not find the task Id in the list of running tasks'
+      });
+      return;
+    }
+
+    this.cleanupRunningState(runningState);
```
jonasfj: I'm curious, but what if it's not in the `list` should try this anyways? or would we just expect that to fail too?

jonasfj:
I think this looks good...

lib/task_listener.js
gregarndt: Good idea.  It's possible that something threw while aquiring the cpu/loopback devices or whatever else that could mean it's not in the list yet as running but still needs things to clean up.

jonasfj:
@gregarndt,
Actually I just realized, how about making `this.pending` a computed property?

You have `this.runningTasks`... So maybe something like:
```js
get pending() {
  return Math.max(this.capacity - this.runningTasks.length, 0);
}
```
See ES6 getter properties, it's quite neat... You can also make it a function, just to make it obvious that it's computed.

Then you won't need increment/decrement, but just send the events necessary in `addRunningTask` and `removeRunningTask`. Then there is less risk that someone calls increment/decrement, and you can accidentally fake high pending by setting it somewhere...

lib/task_listener.js
```diff
@@ -110,21 +108,23 @@ export default class TaskListener extends EventEmitter {
     return cancelListener;
   }
 
+  get availableCapacity() {
+    return Math.max(this.capacity - this.runningTasks.length, 0);
```
jonasfj: Maybe log an `[alert-operator]` if `this.capacity - this.runningTasks.length` is negative... I mean that should never be able to happen, right?

lib/shutdown_manager.js
```diff
@@ -89,9 +89,7 @@ export default class ShutdownManager extends EventEmitter {
     this.taskListener.on('working', this.onWorking);
 
     // Kick off the idle timer if we started in an idle state.
-    if (taskListener.pending === 0) {
-      this.onIdle();
-    }
+    if (taskListener.isIdle()) this.onIdle();
```
jonasfj: nit,
```js
if (taskListener.isIdle()) {
  this.onIdle();
}
```
See mozilla code style guidelines...

lib/task_listener.js
gregarndt: The only time that I can think of this happening (unless something goes greatly wrong) is when capacity is set to 0 when a node is being terminated and anything relying on tasklistener.avaiableCapacity checks it before the running tasks are aborted.

lib/shutdown_manager.js
jonasfj: So I don't care... but coding style guide is unambiguous on the subject:
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style

Then again, I just realized I've consistently done connectives wrong... But at least I know now...

lib/task_listener.js
```diff
@@ -212,24 +198,63 @@ export default class TaskListener extends EventEmitter {
...
+  }
+
+  removeRunningTask(runningState) {
+    let taskIndex = this.runningTasks.findIndex((runningTask) => {
+      return (runningTask.taskId === runningState.taskId) &&
+        (runningTask.runId === runningState.runId)
+    });
+
+    if (taskIndex === -1) {
+      this.runtime.log('[warning] running task removal error', {
```
jonasfj: Are we should this shouldn't be `[alert-operator]`, it seems to me that this breaks an important invariant and we should know if this ever happens...

Mostly because the result might be that `idle` might not be emitted etc... And we don't know what is right to do here...

jonasfj:
Looks good to me... mainly alerting that concerns me...

We should probably develop some form of assert modules (maybe in tc-base) that will log with something like `[alert-operator]` and optionally throw an exception too... It would neat to have... Well, I think we have bug for discussing this...

lib/task_listener.js
jonasfj: ahh, so it can happen...

jonasfj: When I say can happen, I mean it can happen intentionally... So we should obviously not log a warning.... But we might leave a not that this can happen in case of dynamic capacity change.. .Which is used as shutdown... I suspect dynamic capacity change could be used for other things later too... Like test devices that we talked about.

PR #92, opened 2015-05-30
Limit things that output to the log when running tests

gregarndt:


gregarndt:
@jonasfj Outside of some things docker puts in the logs when an image is cached, this seems to remove some of the unnecessary mess you've hate being in the logs :)  Check out the last taskgraph link.  The failure

test/garbage_collection_test.js
```diff
@@ -51,7 +47,7 @@ suite('garbage collection tests', function () {
 
     var gc = new GarbageCollector({
       capacity: 1,
-      log: log,
+      log: debug,
```
jonasfj: you did `var log = debug` above...  but this is fine too... Basically we'll get the logs if we set `DEBUG` correctly right?

jonasfj:
sorry for the late `r++` I love that we can still have all the logs with `DEBUG`...

PR #91, opened 2015-05-29
Cleanup states (allow links to specify env vars)

jonasfj:
* Clean-up how stats are collected for `states.js` (this may have altered the names a bit)
  * Allow `links` hooks to specify environment variables (which may be overwritten by task-specific variables).

I think this should work, I didn't run test yet... Test takes 45min, so I'll do that later as I have more to play with right now...
Perhaps docker-worker CI will pick it up and run the tests for me.  * Clean-up how stats are collected for `states.js` (this may have altered the names a bit)
  * Allow `links` hooks to specify environment variables (which may be overwritten by task-specific variables).

I think this should work, I didn't run test yet... Test takes 45min, so I'll do that later as I have more to play with right now...
Perhaps docker-worker CI will pick it up and run the tests for me.

PR #94, opened 2015-06-01
Docker in Docker for untrusted containers

jonasfj:
Let's land PR #91  first (also PR #93 or this will leak data volumes)... But then this should be a matter of figuring out how to write tests...
and well, just making sure it actually works...

I've tested `dind-service` intensively this weekend... so I'm reasonably confident that it'll work.

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -38,6 +38,7 @@ module.exports = {
   taskclusterLogImage: 'taskcluster/livelog:v3',
   testdroidProxyImage: 'taskcluster/testdroid-proxy:0.0.6',
   balrogVPNProxyImage: 'taskclusterprivate/taskcluster-vpn-proxy:0.0.1',
+  dindImage: 'jonasfj/dind-service:v1.6',
 
   alivenessCheckInterval: 30000, // 30 seconds
 
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -38,7 +38,7 @@ module.exports = {
   taskclusterLogImage: 'taskcluster/livelog:v3',
   testdroidProxyImage: 'taskcluster/testdroid-proxy:0.0.6',
   balrogVPNProxyImage: 'taskclusterprivate/taskcluster-vpn-proxy:0.0.1',
-  dindImage: 'jonasfj/dind-service:v1.6',
+  dindImage: 'taskcluster/dind-service:v1.6',
 
   alivenessCheckInterval: 30000, // 30 seconds
 
--- a/lib/features/dind.js
+++ b/lib/features/dind.js
@@ -50,13 +50,22 @@ export default class DockerInDocker {
       }
     });
 
+    // Terrible hack to get container promise proxy.
+    this.container = docker.getContainer(this.container.id);
+
+    // Start the container
+    await this.container.start({});
+
+    // Find socket path
     let socketPath = path.join(this.tmpFolder, 'docker.sock');
 
     try {
       await waitForSocket(socketPath, INIT_TIMEOUT);
     } catch(err) {
+      debug("Failed to start dind-service");
       throw new Error('Failed to initialize dind-service waiting for socket');
     }
+    debug("dind-service now running!");
 
     return {
       binds: [{
@@ -68,7 +77,7 @@ export default class DockerInDocker {
   }
 
   killed(task) {
-    this.container.stop();
+    this.container.stop(); // don't are to wait for this, GC will kill hard
     task.runtime.gc.removeContainer(this.container.id);
 
     // Remove temporary folder, this should be possible even though contain
@@ -77,7 +86,10 @@ export default class DockerInDocker {
     rmrf(this.tmpFolder, err => {
       // Errors here are not great, but odds of a slugid collision are fairly
       // slim...
-      debug("Failed to remove tmpFolder: %s", err.stack);
+      if (err) {
+        debug("Failed to remove tmpFolder: %s", err.stack);
+      }
     });
   }
 }
+
--- a/lib/features/dind.js
+++ b/lib/features/dind.js
@@ -77,10 +77,9 @@ export default class DockerInDocker {
   }
 
   killed(task) {
-    this.container.stop(); // don't are to wait for this, GC will kill hard
     task.runtime.gc.removeContainer(this.container.id);
 
-    // Remove temporary folder, this should be possible even though contain
+    // Remove temporary folder, this should be possible even though container
     // is running... Well, maybe docker will complain that mounted folder is
     // deleted (delete socket while running is not a problem)
     rmrf(this.tmpFolder, err => {
--- a/lib/features/dind.js
+++ b/lib/features/dind.js
@@ -1,0 +1,83 @@
+/**
+ * This module handles the creations of a dind-service container, links and sets
+ * the DOCKER_HOST environment variables such that task container will be able
+ * to use `docker`. For security reasons you can't run privileged containers,
+ * use capabilities or volume mount host folders. But it should work out of the
+ * box for common docker flows.
+ */
+
+import _ from 'lodash';
+import waitForSocket from '../wait_for_socket';
+import { pullImageStreamTo } from '../pull_image_to_stream';
+import slugid from 'slugid';
+import rmrf from 'rimraf';
+import path from 'path';
+import fs from 'fs';
+let debug = require('debug')('docker-worker:features:dind');
+
+// Maximum time to wait for dind-service to be ready
+const INIT_TIMEOUT = 30 * 1000;
+
+export default class DockerInDocker {
+  constructor() {
+    // dind-service container
+    this.container = null;
+    this.tmpFolder = path.join('/tmp', slugid.v4());
+  }
+
+  async link(task) {
+    let docker = task.runtime.docker;
+
+    // Pull docker image and create temporary folder
+    await Promise.all([
+      pullImageStreamTo(docker, task.runtime.dindImage, process.stdout),
+      new Promise((accept, reject) => {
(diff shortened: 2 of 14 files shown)


jonasfj:
Rebased on fixed PR #91.

lib/states.js
```diff
@@ -29,50 +29,75 @@ Each "state" in the lifetime is a well defined function:
...
+    {
+      links: [
+        {name: 'container name', alias: 'alias in task container'}
       ]
+      env: {
+        'name-of-env-var':  'value of variable'
+      },
+      binds: [
+        {source: '/path/on/host', target: '/path/in/container', readOnly: true}
+      ]
```
jonasfj: I think collecting `links`, `env` and `binds` from features in the `link` hook makes sense...

We can now move "cache folders" and "encrypt env vars" into features... which would dramatically simplify the `task.js` file. And it would allow us to experiment with the rewrite/redesign of cache folders we've been talking about (preloaded read-only caches and all the other variants).

jonasfj:
@gregarndt, I'm not sure what I did to make it mad here but... something is funny... it's like it doesn't want to install...

config/defaults.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
   taskclusterLogImage: 'taskcluster/livelog:v3',
   testdroidProxyImage: 'taskcluster/testdroid-proxy:0.0.6',
   balrogVPNProxyImage: 'taskclusterprivate/taskcluster-vpn-proxy:0.0.1',
+  dindImage: 'jonasfj/dind-service:v1.6',
```
gregarndt: Will we move this to the taskcluster project on docker hub?

jonasfj: yeah, we should push it to `taskcluster/dind-service:v1.6` I didn't have access at the time.

lib/wait_for_socket.js
```diff
@@ -0,0 +1,30 @@
+var Promise = require('promise');
```
gregarndt: looks like you converted this to using new import and didn't get rid of previous requires.

lib/wait_for_socket.js
```diff
@@ -0,0 +1,30 @@
...
+var fs = require('fs');
+import Promise from 'Promise';
```
gregarndt: I think this should be from 'promise' right?

test/images/dind-test/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM        gliderlabs/alpine:3.2
```
gregarndt: We probably could add the building of this image to https://github.com/taskcluster/docker-worker/blob/master/build.sh

lib/wait_for_socket.js
jonasfj: yeah, I wonder how this worked... :)

test/integration/dind_test.js
```diff
@@ -0,0 +1,41 @@
+
```
gregarndt: nit: leading blank line

test/integration/dind_test.js
```diff
@@ -0,0 +1,41 @@
...
+    if (worker) {
+      await worker.terminate();
+      worker = null;
+    }
+  });
+
+  test('run docker in docker', async () => {
+    let result = await worker.postToQueue({
+      payload: {
+        image: 'jonasfj/dind-test:v1',
```
gregarndt: just a reminder to update this to be taskcluster/ once we push it there

test/integration/dind_test.js
```diff
@@ -0,0 +1,41 @@
...
+        command: [''],
+        features: {
+          bufferLog: false,
```
gregarndt: What is bufferlog?

test/integration/dind_test.js
```diff
@@ -0,0 +1,41 @@
...
+          azureLiveLog: false,
```
gregarndt: this is false by default

lib/features/dind.js
```diff
@@ -0,0 +1,95 @@
...
+const INIT_TIMEOUT = 30 * 1000;
+
+export default class DockerInDocker {
+  constructor() {
+    // dind-service container
+    this.container = null;
+    this.tmpFolder = path.join('/tmp', slugid.v4());
+  }
+
+  async link(task) {
```
gregarndt: Do we want to scope this feature at all?

lib/wait_for_socket.js
gregarndt: Not sure how it worked for you, but I initially noticed it because it broke all the CI tests :)

test/images/dind-test/Dockerfile
jonasfj: Yay, that is awesome :)

lib/features/dind.js
```diff
@@ -0,0 +1,95 @@
...
+        readOnly: true
+      }]
+    };
+  }
+
+  killed(task) {
+    this.container.stop(); // don't are to wait for this, GC will kill hard
+    task.runtime.gc.removeContainer(this.container.id);
+
+    // Remove temporary folder, this should be possible even though contain
```
gregarndt: such a huge nit! "contain" should be "container" :)

test/integration/dind_test.js
jonasfj: done,

lib/features/dind.js
```diff
@@ -0,0 +1,95 @@
...
+      binds: [{
+        source: socketPath,
+        target: '/var/run/docker.sock',
+        readOnly: true
+      }]
+    };
+  }
+
+  killed(task) {
+    this.container.stop(); // don't are to wait for this, GC will kill hard
```
gregarndt: "don't are to wait for this" ?

lib/features/dind.js
```diff
@@ -0,0 +1,95 @@
...
+    task.runtime.gc.removeContainer(this.container.id);
+
+    // Remove temporary folder, this should be possible even though contain
+    // is running... Well, maybe docker will complain that mounted folder is
+    // deleted (delete socket while running is not a problem)
+    rmrf(this.tmpFolder, err => {
```
gregarndt: yea this should be good enough for now, but what I would like with the GC write is when marking a container for remove, it has things it needs to clean up, including volumes it has lying around.

lib/features/dind.js
jonasfj: oh, yeah, I considered it when I started... Now I can't decide...
Technically it's locked down, so there not really any reason to scope it.

However, I suspect we'll add scopes when we make dind-cache folders... But then we might want to allow things run without persistent cache folders.

Only reason people shouldn't use this all the time is that pulling down docker images is not reliable...

lib/features/dind.js
jonasfj: maybe for this we should data-volumes...

So this tmpfolder only holds the domain socket... just fyi,

lib/features/dind.js
jonasfj: Not even sure why I call stop here...

test/integration/dind_test.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+  });
+
+  test('run docker in docker', async () => {
+    let result = await worker.postToQueue({
+      payload: {
+        image: 'taskcluster/dind-test:v1',
+        command: [''],
+        features: {
+          bufferLog: false,
```
gregarndt: Hrm, thought I left a comment here, but github isn't showing it.  Not sure what "bufferLog" is

test/integration/dind_test.js
```diff
@@ -0,0 +1,40 @@
...
+          azureLiveLog: false,
```
gregarndt: this is false by default

test/integration/dind_test.js
```diff
@@ -0,0 +1,40 @@
...
+      await worker.terminate();
+      worker = null;
+    }
+  });
+
+  test('run docker in docker', async () => {
+    let result = await worker.postToQueue({
+      payload: {
+        image: 'taskcluster/dind-test:v1',
+        command: [''],
```
gregarndt: Would it be clearer in the test what's going on here by calling docker run here instead of in an entrypoint script?

gregarndt: This could also make it so we could test building images or what not.

gregarndt:
This looks good, the taskgraph is green (other than the two known failures on chunk 4).  There's just some comments. r+

lib/task.js
```diff
@@ -318,7 +339,11 @@ export default class Task {
     ));
 
     try {
-      await this.runtime.stats.timeGen(`tasks.time.states.${stat}.killed`, this.states.killed(this));
+      // Run killed hook and record reason why we aborted
+      await this.runtime.stats.timeGen(
```
gregarndt: Isn't the stat recorded by the _invoke method in states.js?

test/integration/dind_test.js
jonasfj: Oh, it probably doesn't exist anymore...

test/integration/dind_test.js
jonasfj: Agree... we can test more later... It's hard to stuff things into command, but this one is so simple...

lib/task.js
jonasfj: Nope... you have a stat for each reason it was aborted... that is not recorded by `_invoke`...

This changes the layout of your stats, but it'll now be `tasks.time.states.killed-by-abort.${stat}` in addition to ``tasks.time.states.killed`.

I believe the `stat` variable is some sort of reason... :)

test/integration/dind_test.js
jonasfj: For the record, this is fixed... I just move the blank from top of the file down here... in github diff :)

PR #96, opened 2015-06-02
Devices should influence overall avaialble capacity

gregarndt:


lib/devices/audio_device_manager.js
```diff
@@ -42,19 +42,29 @@ export default class AudioDeviceManager {
   }
 
   getAvailableDevice() {
-    debug('Aquiring available device');
-    for (let device of this.devices) {
-      if (device.active) continue;
-      device.aquire();
-      debug(`Device: ${device.path} aquired`);
-      return device;
+    let devices = this.getAvailableDevices();
```
jonasfj: I think you have a case of infinite recursion here...

lib/devices/audio_device_manager.js
```diff
@@ -42,19 +42,29 @@ export default class AudioDeviceManager {
...
+      throw new Error(`
+        Fatal error... Could not acquire audio device:
+
+        ${JSON.stringify(this.devices)}
+      `);
     }
 
-    throw new Error(`
-      Fatal error... Could not aquire audio device:
+    debug('Acquiring available device');
```
jonasfj: empty line why? just curious if it's intentional maybe use `\n`.

lib/devices/audio_device_manager.js
jonasfj: nvm...

lib/devices/phone_device_manager.js
```diff
@@ -0,0 +1,78 @@
...
+
+  async devices() {
+    let devices = await this.client.getDevices(this.deviceFilter);
+    let deviceList = devices.map((device) => {
+      return new Phone(device);
+    });
+
+    debug(`List of ${deviceList.length} phones created`)
+    return deviceList;
+  }
```
jonasfj: So won't all workers have a same list of devices?

I feel like it's bad idea to recreate the `Phone` objects... maybe it would be smarter to maintain a  map from `phone.id` to phone object... and then add phone devices that shows up, and remove phone devices that disappears from the API...

I argue for this mostly because having an object representing a phone isn't super fun, if there is two instances of the object representing the same phone device. And `getAvailableDevice` is probably used by some consumer that keeps the `Phone` object around until the task is done.

gregarndt: So ideally this list of devices will be devices the worker currently has sessions for. It will create the sessions and then base claim capacity on how many phones for sure it has for tasks. 

Something better could be done here.  Right now it's only used for getting a list of online devices and once bitbar finishes some Api work we'll be able to create sessions before claiming. That part I'm still waiting on from them.

lib/task_listener.js
```diff
@@ -39,6 +30,12 @@ export default class TaskListener extends EventEmitter {
     this.runningTasks = [];
     this.taskQueue = new TaskQueue(this.runtime);
     this.taskPollInterval = this.runtime.taskQueue.pollInterval;
+
+    if (this.runtime.isolatedContainers) {
```
jonasfj: Curious, can't we just remove the `isolatedContainers` option and activate it with `deviceManagement.cpu.enabled = true` ?

I might not be up to date with how config is loaded from user-data, so if `deviceManagement` can't fit in user-data forget it :)
Just and idea...

lib/task_listener.js
```diff
@@ -113,26 +110,51 @@ export default class TaskListener extends EventEmitter {
...
     // negative number, hence the use of at least returning 0.
-    return Math.max(this.capacity - this.runningTasks.length, 0);
+    let deviceCapacity;
+    try {
+      deviceCapacity = await this.deviceManager.getAvailableCapacity();
+    }
+    catch (e) {
+      // If device capacity ccannot be determined for device managers configured
+      // for the worker, then default to 0
+      debug('Error determining device capacity');
```
jonasfj: Maybe alert with `[alert-operator]` this is pretty bad right?

Capacity zero is not the problem... the problem is that we failed to determine capacity, that shouldn't happen am I right, if so how?

lib/task_listener.js
gregarndt: Ah. We could. There were some code changes I was making when coming up with devicemanagement and didn't update this: it can fit in user data. Or at least it's working now for me

jonasfj: Hmm... I see how `isolatedContainers` might make more sense longer down...

Maybe, just set `deviceManagement.cpu.enabled = false` in `config/defaults.js`, then the code here only needs to be `deviceManagement.cpu.enabled = true`, which is a bit less obscure than the object assignment...

Maybe, not sure... probably just a detail.

lib/task_listener.js
gregarndt: it could happen when the testdroid client throws an error when the api is unreachable or returns an error.  Perhaps the client itself shouldn't throw errors.

lib/task_listener.js
```diff
@@ -113,26 +110,51 @@ export default class TaskListener extends EventEmitter {
...
+        }
+      );
+    }
+
+    return hostCapacity;
   }
 
   async getTasks() {
-    if (this.availableCapacity === 0)  return;
+    let availableCapacity = await this.availableCapacity();
```
jonasfj: Can we reliably call bitbar every 5s... If not maybe the phone manager needs some sort of rate limiter to make sure that it doesn't actually call more than say once a minute...

gregarndt: I was going to set task polling to a longer interval for these workers. Like 1 minute.

config/defaults.js
```diff
@@ -1,4 +1,26 @@
...
+    // and that session passed into the task.
+    // Example:
+    //   {
+    //     enabled: true,
+    //     'sims': '1',
+    //     'type': 'flame-kk'
+    //   }
+    'phone': {
+      enabled: false
+    }
```
jonasfj: Why not add defaults, or at least just list the properties illustratively (and add `maxUpdateInterval`).
```js
{
  enabled: false,
  sims: '1', // number of sim cards
  type:  undefined, // device type
  maxUpdateInterval: 60 * 1000 // max time between updating list of devices..
}
```

lib/devices/phone_device_manager.js
```diff
@@ -0,0 +1,78 @@
...
+        config.testdroid.password
+    );
+    this.deviceFilter = {
+      'type': config.deviceManagement.phone.type,
+      'sims': config.deviceManagement.phone.sims
+    };
+  }
+
+  async devices() {
+    let devices = await this.client.getDevices(this.deviceFilter);
```
jonasfj: I suspect it's best not to poll this every 5s, so maybe implement `maxUpdateInterval` as suggested above...

And just return the previous list without modifications if `maxUpdateInterval` hasn't passed... Ie. it's probably a time that you have to count using Date.now()...

Note, if the list of devices contains available devices we have a problem...
 1. we can't really update every 5s to see if it's still available.
 2. we can't really assume it's still available.

For now, this might not be a problem... because each worker can handle ~10 devices (if I recall you correctly), so we're not going to have many works, hence, polling every 5s might work.

lib/task_listener.js
gregarndt: and there's no need to hit the API at all once we have enough locked sessions for each task the work could claim. We would only need to hit their API once we lose a device from that pool or need to refresh a session.

jonasfj:
Generally this looks decent to me... I might have implement base classes for `DeviceManager` and `Device` such that we could have default implementations of methods that throws a not-implemented error..
Also it's easier to check the base class to see what methods are required to implement, and documenting how the class is intended to be consumed in the base class is also nicely abstract...

Anyways, that's a nit thing..
---

My only slight concern here is polling bitbar every 5s from multiple workers... Especially if/when we scale this might be hard for them to handle.

lib/task_listener.js
gregarndt: Of course that means nothing until we actually start maintaining locked device sessions prior to claiming tasks. This was the ground work to be a best effort at determining available devices and would be improved once we can create those sessions

jonasfj:
Oh, realize it's not clear `r+`, if the bitbar polling every 5s isn't an issue.

gregarndt:
Absolutely agree about 5s. I was going to set task polling to 1 minute which would accomplish the similar affect of your maxupdateinterval suggestion I think.

Definitely like the idea of a base class for device. Will look into that. Thanks for the comments!

lib/task_listener.js
gregarndt: Ok, so alert-operator has been added, and we can set the treshold in papertrail for how often it notifies us, thanks!

lib/task_listener.js
jonasfj: Ahh, that makes a lot of sense.. Sounds good to me..

lib/task_listener.js
jonasfj: Yeah, alternative would be logging it to stats someday... Not important..

jonasfj:
Sounds like the `5s` polling thing was thought of already... So I suggest you land it. `r+` :)

PR #97, opened 2015-06-04
Update testdroid proxy with timeout and remove vpn

gregarndt:
Remove the need for providing vpn credentails now that us-west-1 is VPC peered with bitbar.  Also started with a copy of eslintrc from provisioner

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,42 @@
+{
+  // https://medium.com/@dan_abramov/lint-like-it-s-2015-6987d44c5b48
+  "parser": "babel-eslint",
+
+  "env": {
+    "node": true,
+    "mocha": true,
+    "es6": true,
+  },
+
+  "rules": {
+    "max-len": [2, 120, 2],
+    "quotes": [2, "single"],
+    "complexity": [1, 10],
+    "curly": 2,
+    "default-case": 2,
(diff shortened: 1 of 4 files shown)


gregarndt:
@jonasfj Mind giving this a quick sanity check?

.eslintrc
```diff
@@ -0,0 +1,42 @@
...
+    "semi": [2, "always"],
+    "brace-style": [2, "1tbs", { "allowSingleLine": true }],
+    "default-case": 2,
+    "no-eq-null": 2,
+    "no-floating-decimal": 2,
+    "no-self-compare": 2,
+    "no-throw-literal": 2,
+    "radix": 2,
+    "yoda": [1, "never"],
+    "strict": [2, "global"],
```
jonasfj: No need for `use strict` with babeljs..

Perhaps we need to steal some rules from: https://github.com/babel/babel/blob/master/.eslintrc

gregarndt: ah, I'll update it, this was just a copy from the provisioner I think.

jonasfj:
Looks solid to me.. Feel free to ignore comment on linter, I assume choice of rules is work in progress :)

`r+`, but check that image id doens't need update.. Merge anytime.

PR #99, opened 2015-06-05
Rename tests that didn't have _test suffix

gregarndt:
This fixes two tests that were not included in the ci tests.

Also fixed an issue where encrypted environment variables were not included in the container config.


gregarndt:
@jonasfj Care to give this a quick review?  Two tests were mistakenly not named correctly and so it didn't catch a bug we introduced with a recent merge.  Fortunately this wasn't rolled out anywhere.  The first task graph shows the failure for encrypted env variables, the second graph should show that it's fixed (the test is in chunk 2).

lib/task.js
```diff
@@ -183,13 +183,14 @@ export default class Task {
   */
   async dockerConfig(linkInfo) {
     let config = this.task.payload;
-    // Allow task specific environment vars to overwrite those provided by hooks
-    let env = _.defaults({}, config.env || {}, linkInfo.env || {});
 
     await this.runtime.privateKey.decryptEnvVariables(
       this.task.payload, this.status.taskId
```
jonasfj: instead of giving `this.task.payload` can we give it `config` so that it's less confusing...

It's the same object I see, but there is a lot of room of doing it wrong right here :)

jonasfj:
`r+` feel free to merge.

lib/task.js
gregarndt: yea, I can change that.  What I actually want to do (but won't in this PR) is to pass it the payload, but it returns the envs back rather than doing magical things and modifying task.payload that is then used later.

gregarndt: fixed in 7ee4b06

PR #100, opened 2015-06-07
Support device capabilities specified in task payload

gregarndt:
Phone device capabilities should be specified in the task.payload.capabilities.devices.phone object.

Right now these properties are just injected into the task container as an env variable until sessions/flashing is done outside of the task.


jonasfj:
@gregarndt, Looks good to me `r+`.

PR #101, opened 2015-06-08
Convert statsd to influx

gregarndt:


gregarndt:
@jonasfj You mind taking a look at this?  Pretty much this should convert us over to influx.

lib/stats/series.js
```diff
@@ -0,0 +1,209 @@
...
+  })
+});
+
+export const workerHDTotal = new base.stats.Series({
+  name: 'worker.hd_total',
+  columns: Object.assign({}, BASE_WORKER_SCHEMA, {
+    // Time when worker received shutdown signal
+    value: base.stats.types.Number
+  })
+});
```
jonasfj: I know that this is what influxdb teaches us to do...
But I'm tempted to merge `workerHDUsed`, `workerHDFree` and `workerHDTotal` and report them:
```js
export const workerHD = new base.stats.Series({
  name: 'worker.hd',
  columns: Object.assign({}, BASE_WORKER_SCHEMA, {
    used: base.stats.types.Number,
    free: base.stats.types.Number,
    total: base.stats.types.Number
  })
});
```
This certain isn't what the 0.9 docs advice, but I wonder if it's wrong??? This way we can ask relative questions like how many have `< 50%` free, etc... Having free space without knowing the size of the disk isn't very fun... Same with `used`. I suspect that don't need both `used` and `free`, but maybe `size`, `used` and `freePerCapacity`.

Also I think you can save this as bytes, `2^53` (max in js), can count to 9 Peta bytes... So counting in GB might also we be worth it, either way comment should reflect this.. btw, I like comments, especially when they aren't duplicates :)

gregarndt: I was tempted to do that, but then decided to follow the crowd and store one value per series :).

Can we not do queries across series?

gregarndt: whoops, I'll remove those wrong comments...stupid copy/paste

lib/stats/series.js
```diff
@@ -0,0 +1,209 @@
+import base from 'taskcluster-base';
+
+const BASE_WORKER_SCHEMA = {
+    workerId: base.stats.types.String,
+    workerType: base.stats.types.String,
+    workerGroup: base.stats.types.String,
+    provisionerId: base.stats.types.String
+};
```
jonasfj: I suspect we want to add `capacity` here, possibly also `instanceType`... Most cloud providers have a concept of `instanceType`, most even call it `instanceType`. We can just configure the default `instanceType` as `Unknown` if we don't know any.

gregarndt: Ok, I can definitely add that.  That was one of the reasons I abstracted this out so it was easy to change and report it for all metrics.

lib/stats/series.js
jonasfj: No we efficiently query cross series... If say we wanted to look at < 50% free disk space...
I'm not sure how fast that would be in the first place, but it certainly would be very very slow, if they are in different series.

Note, I'm using the multiple values in single series for other components. 

----
Actually, I'm note sure... 

But [let's avoid `.` in series names so we don't have to quote them in queries](http://influxdb.com/docs/v0.9/concepts/schema_and_data_layout.html)...

jonasfj:
I'll take a look at this a bit later... I'm not done... In the meantime think about or read docs on whether we should one value per series. I'm reasonably sure we should just do more than one, but it's possible that we should ask the influxdb people (what's the point of support for multiple fields if you're only supposed to use one) :)

bin/worker.js
```diff
@@ -156,27 +158,22 @@ co(function *() {
...
+  config.stats.record('workerStart', Date.now()-os.uptime() * 1000);
+
+
+  // Only catch these metrics when running on aws host.  Running within
+  // test environment causes numerous issues.
+  if (host === 'aws') {
+    base.stats.startProcessUsageReporting({
+      drain: config.stats.influx,
+      component: 'docker-worker',
+      process: 'docker-worker'
```
jonasfj: Hmm... we should modify this in `base` to take `tags` like the thing I suggested for in the PR about client API stats...

config/defaults.js
```diff
@@ -150,10 +150,12 @@ module.exports = {
...
-  // Statsd configuration options (these are totally optional).
-  statsd: {
-    prefix: process.env.STATSD_PREFIX || '',
-    url: process.env.STATSD_URL || 'tcp://localhost:8125'
+  influx: {
+    connectionString: undefined,
+    maxDelay: 10,
+    maxPendingPoints: 100,
+    // Only collect host level metrics every minute
+    hostMetricsInterval: 60 * 1000
```
jonasfj: I'm tempted to suggest that we put this else where... all the other config options here fit right into `base.stats.Influx` constructor...

lib/queueservice.js
```diff
@@ -107,12 +107,10 @@ export default class TaskQueue {
     let claim;
     try {
       claim = await this.stats.timeGen(
-        'tasks.time.claim',
+        'claimTask',
```
jonasfj: I don't like this... Let's do that with the client API stats handler... See two PRs filed against you :)

IMO if we want to record statistics about a method I see three options:
 1. Add stats recording (or callback/event) from with method
 2. Monkey patch the method to record stats
 3. Wrap the entire thing the method lives on and add stats recording there

Adding stats at call-sites throughout the code is bad. It reduces readability and adds things we don't want to see when we review logic. It's also error prone, let's strive to only use it in corner cases (yeah, we have a lot of those :))

lib/stats/stat.js
```diff
@@ -0,0 +1,106 @@
...
+    this.influx = new base.stats.Influx(config.influx);
+    this.baseWorkerStats = {
+      workerId: config.workerId,
+      workerGroup: config.workerGroup,
+      workerType: config.workerType,
+      instanceType: config.workerNodeType,
+      provisionerId: config.provisionerId,
+      capacity: config.capacity
+
+    };
```
jonasfj: I  see how this makes sense... How about we modify `base.stats.Series#reporter`, such that we can do the following:
```js
var mySeries = new base.stats.Series({.....});
var drain = new base.stats.Influx({....});

var reportMySeries = mySeries.reporter({
  drain: drain,
  tags: {
    workerId: '...',
    workerGroup: '...',
    ....
  }
});
```

Part of the idea with `Series` was that things could be more typed, instead of using strings as identifiers, which is what the `record` method does here. Hence instead of `stats.record('mySeries', value)` we do `stats.reportMySeries({value: ....})`.

Adding this to `base.Series` should be easy... ping me if you like the idea... I think I can already see one, possibly two, uses cases for this in `base.stats` already...

lib/task_listener.js
```diff
@@ -341,5 +341,7 @@ export default class TaskListener extends EventEmitter {
         });
       }
     }
+
+    this.runtime.stats.record('taskError');
```
jonasfj: Is this iike a JS error... In that case I suspect you could add a tag/string `err.code || err.name || 'Error'` it typically has low carnality and is a good way to see what errors plague us the most..

jonasfj:
So I think we can simplify some things by moving a little into `base.stats` in particular the reporter...

I would also like to discuss merging the series... We loose a lot of information by having separate series for each measure (ie. exit code and runTime would be nice together). Same goes for many other things, like free disk space and total disk space, or free memory and total memory.

I'm very interested in updating `base.stats`...


Anyways, I see nothing bad here, I'm sure you can land it, but I would probably look at moving something into `base.stats` and merge series to make it more usable.

lib/queueservice.js
gregarndt: I'll be taking a look at those PRs this morning.  What I'm wanting to capture (and probably can with those PRs) is worker level claims/failures/exceptions as well as the time to claim the task (including all retries, etc).  My hopes is that the client can give me that and then I'm definitely all for not including those things here.

lib/stats/stat.js
gregarndt: I definitely like the idea of adding tags when the series is created that will be added to all points recorded.  It makes a lot of sense .  Would simplify a bit in the provisioner as well where there are multiple things recorded with each series https://github.com/taskcluster/aws-provisioner/blob/master/provisioner/influx-series.js

Also, the purpose of the Stat class was not only to include certain columns with each series, but also not do things like:
https://github.com/taskcluster/aws-provisioner/blob/master/provisioner/aws-manager.js#L56-62

I figured calling stats.record(<series name>..) was simpler to read, but I might be mistaken.

lib/task_listener.js
gregarndt: Yea, this is to catch things that bubble up and were not caught elsewhere or when getting a device or claim status throws an exception at the task_listener.js level.  We can add something here to log something.

gregarndt:
So I have merged memory and HD into their own series that have columns like free, total, used.

lib/queueservice.js
gregarndt: so far, the stats added to the client seem to be exactly what we're going after, still reviewing, but I think once the new client lands we could change these a bit

lib/stats/stat.js
jonasfj: @garndt,
there is next to no overhead in calling `MySeries.reporter(drain, tags)`, you could put it behind a `getter` but it would be way overkill...

Look at the code in base, it's just a quick check (up front) and a tiny bit of memory.

lib/stats/stat.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * Return a reporter for a given series.
+   *
+   * @param {String} seriesName - name of the series
+   *
+   * @returns {Series} series reporter
+   */
+  getReporter(seriesName) {
```
jonasfj: This is total overkill... In the constructor just do:
```js
_.forIn(series, (serie, name) => {
  this[report + name] = series.reporter(this.influx, this.baseWorkerStats);
});
```

Then you'll have all the series initialized initially, and if there is any bugs they'll throw early.
Also the overhead to calling `.reporter()` is very small look at it... It just keeps a reference to `baseWorkerStats` and returns a closure... No real memory allocation of significance. Only a quick few sanity checks.

lib/stats/host_metrics.js
```diff
@@ -0,0 +1,49 @@
...
+  stats.record('workerCPULoad', (loadTime/totalCPUTime) * 100);
+
+  diskspace.check(dockerVolume, function (err, total, free) {
+    stats.record('workerHD', {
+      free: free,
+      used: total-free,
+      total: total
+    });
+  });
+}
```
jonasfj: So I don't know about you... But I'm tempted to merge both memory, disk and uptime related stats...

Is this crazy? I'm thinkin queries like `median(freeDisk) where uptime > 48 hours` could tell us if we have GC issues, or memory leaks. Also could identify if nodes gets worse over time... basically..

jonasfj: Not, sure this is sane... Just pointing out an opportunity.

lib/stats/series.js
```diff
@@ -0,0 +1,195 @@
...
+  columns: Object.assign({}, BASE_WORKER_SCHEMA, {
+    type: base.stats.types.String,
+    sims: base.stats.types.String
+  })
+});
+
+export const statesLink = new base.stats.Series({
+  name: 'states_linked',
+  columns: Object.assign({}, BASE_WORKER_SCHEMA, {
+    value: base.stats.types.Number
```
jonasfj: Please, let's merge these into one series...

And lets call `value` for `duration`?

I see one series called `stateChanged` with columns `duration` and `state` as string (later tag in 0.9)..

lib/stats/series.js
```diff
@@ -0,0 +1,195 @@
...
+  name: 'tasks_error',
+  columns: Object.assign({}, BASE_WORKER_SCHEMA, {
+    value: base.stats.types.Number
+  })
+});
+
+export const runTimeExceeded = new base.stats.Series({
+  name: 'tasks_run_time_exceeded',
+  columns: Object.assign({}, BASE_WORKER_SCHEMA, {
+    value: base.stats.types.Number
```
jonasfj: How about we have one extra field here `maxRunTime` and `count: 1` which is always `1` because each observation is one...

jonasfj:
`r+`

lib/shutdown_manager.js
```diff
@@ -10,12 +10,16 @@ export default class ShutdownManager extends EventEmitter {
...
     this.config = config;
+    this.stats = config.stats;
     this.nodeTerminationPoll = config.shutdown.nodeTerminationPoll || 5000;
     this.onIdle = this.onIdle.bind(this);
     this.onWorking = this.onWorking.bind(this);
   }
 
   async shutdown() {
+    this.stats.record('workerShutdown', Date.now());
+    this.stats.close();
```
jonasfj: Please do an `await` here...

jonasfj: Or do the await in parallel with this.taskListener.close() using Promise.all

PR #102, opened 2015-06-12
Fix bug 1174035

jonasfj:
Wait for all promises no matter how many errors we have...


jonasfj:
Wait for tests... I didn't run any :)

PR #103, opened 2015-06-12
Retry docker image pulls with exponential backoff

gregarndt:
Deployed this to the raptor worker type which was hitting this frequently.
Here is a second attempt made at downloading the livelog image and it succeeded:  https://papertrailapp.com/systems/92517734/events?centered_on_id=546077531615571975

### DIFF
--- a/lib/pull_image_to_stream.js
+++ b/lib/pull_image_to_stream.js
@@ -2,6 +2,8 @@ import DockerImage from './docker_image';
 import dockerUtils from 'dockerode-process/utils';
 import Debug from 'debug';
 import { scopeMatch } from 'taskcluster-base/utils';
+import waitForEvent from './wait_for_event';
+import sleep from './util/sleep';
 
 let debug = Debug('pull_image');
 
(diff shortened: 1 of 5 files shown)


lib/pull_image_to_stream.js
```diff
@@ -14,17 +16,48 @@ export const IMAGE_ERROR = 'Pulling docker image "%s" has failed this may indica
...
-    downloadProgress.pipe(stream, { end: false });
-    downloadProgress.once('error', reject);
-    downloadProgress.once('end', accept);
-  });
+    downloadProgress.pipe(stream, {end: false});
+    downloadProgress.once('error', err => { error = err; });
+    await Promise.race([
+      waitForEvent(downloadProgress, 'error'),
+      waitForEvent(downloadProgress, 'end')
+    ]);
```
jonasfj: I don't like waitForEvent... how about:
```js
try {
    downloadProgress.once('end', accept);
    downloadProgress.once('error', reject);
  });
  return;
} catch (err) {
  if (attempts === maxAttempts) {
    ....
  }
  ....
}
```

jonasfj: Then you can drop the `error` variable and `downloadProgress.once('error', err => { error = err; });` too...

lib/pull_image_to_stream.js
```diff
@@ -14,17 +16,48 @@ export const IMAGE_ERROR = 'Pulling docker image "%s" has failed this may indica
...
+
+    if (!error) {
+      return;
+    }
+
+    if (attempts === maxAttempts) {
```
jonasfj: `attempts >= maxAttempts` just for robustness :)

lib/pull_image_to_stream.js
```diff
@@ -14,17 +16,48 @@ export const IMAGE_ERROR = 'Pulling docker image "%s" has failed this may indica
...
+    }
+
+    let delay = Math.pow(2, attempts - 1) * delayFactor;
+    delay = delay * (Math.random() * 2 * randomizationFactor + 1 - randomizationFactor);
+    debug(
+      `pull image failed ` +
+      `Next Attempt in: ${delay} ms` +
+      `Image: ${image} ` +
+      `Error: ${error} `
+    );
```
jonasfj: Don't use `debug` like this... it's much cheaper to do: `debug("pull image failed Next Attempt %s ... image: %s", delay, image, ...)` Also it's a good idea to dump `err.stack` along with `err` using `%j` so you get any properties set on error.

Example,
```js
debug('pull image failed next attempt in %s ms, Image: %s, err: %s, as JSON: %j', delay, image, err, err, err.stack);
```

The last `err.stack` will be added to it as string... this works well for me...

lib/pull_image_to_stream.js
gregarndt: ah, good idea, I like that much more.

lib/pull_image_to_stream.js
gregarndt: If there is a potential efficiency to doing this, then I'm up for it.  I'll make the chnage.  I am curious though if you know of any documentation about the gains achieved by not using the string templating

jonasfj: String templating compiles to `"mystring " + myvar + " another string"`, this expression is given as argument to `debug()`, hence, following common function call semantics the expression `"mystring " + myvar + " another string"` will be evaluated before the function `debug` is invoked with it as argument.

If you call `debug("my string %s another string", myvar)` there is no expressions to evaluate prior to invocation of `debug`. In the case where `DEBUG` doesn't activate the `debug` function for this file, the result is that `debug = function() {}`. Hence, the string interpolation doesn't take place, if `DEBUG` disables the debug logging, and the call to `debug` has no overhead (a good JS compiler will inline it and dead code elimination may remove it, but even if not overhead is very small). 

On the other hand, if you call `debug("mystring " + myvar + " another string")` where `debug` is an empty function. The JS _may_ be able to recognize that the function `debug` is empty and the expression `"mystring " + myvar + " another string"` has no side effects and then remove it. But this is very very hard. I seriously doubt JS compilers are that great.

However, when using `debug` for string interpolation, even a naive non-optimizing JS interpreter would be able to benefit. String operations are fairly heavy, so being able to disable them when not needed is great. It means that we don't have to feel bad about adding an excessive amount of `debug` statements (as long as we disable them at run time).

Note, using `debug` for string interpolation starts becoming increasingly important if dumping json, as `debug("%j", data)` won't call `JSON.stringify` unless `debug` is activated.

gregarndt: Awesome, I feel +10 smarter now, thanks!

PR #104, opened 2015-06-14
Bug 1174562 - Ensure image is downloaded

gregarndt:
Noticed that when simultaneous downloads were executed by a worker (capacity > 1) sometimes n-1 of n tasks would fail and only one succeed.  The failed tasks all failed when a 404 when pulling the live log image and then trying to create a container based on that image.  The image wasn't available because the one task that didn't fail yet, was retrying to pull the image because it received a 504 when pulling a layer.

What seems to be happening is that the other tasks were waiting for the client to finish pulling, and when the 504 was raised for the task pulling the image, the stream ended for the other tasks and they assumed the image was downloaded and tried to create a container.  Of course the image wasn't there yet so this should provide some sanity check to see if the image is available after the stream ends if no error is raised.

### DIFF
--- a/lib/pull_image_to_stream.js
+++ b/lib/pull_image_to_stream.js
@@ -41,6 +41,18 @@ export async function pullImageStreamTo(docker, image, stream, options={}) {
         downloadProgress.once('error', reject);
         downloadProgress.once('end', accept);
       });
+
+      // Ensure image downloaded after pulling. This is mostly for multiple tasks
+      // pulling at the same time, only one will pull the image while the others wait.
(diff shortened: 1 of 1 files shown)


jonasfj:
`r+` awesome to have this check... even if it's just a sanity check...

We have so many docker issues... that one extra check is lovely.

gregarndt:
Been running this as a prototype on some workers without issue.  Only task aborts I've seen are because of runtime exceeded. should be good, definitely haven't seen the 404 errors I saw yesterday

PR #105, opened 2015-06-15
Bug 1174559 - Make taskcluster log errors be parseable

gregarndt:


jonasfj:
Looks good to me `r+`

---
I wish that `logFmt` was called `logMsg` and we had another one called `logErr` which both did the formatting and did `stream.write`. It's a minor refactor, but most of the methods that use `logFmt` might aswell be writing directly to `stream`...

jonasfj:
Note, `logFmt` is something we can do any day.. Preferably as a separate PR :)

gregarndt:
yea, definitely something that should be done, but could be in another PR.  Thanks!

PR #106, opened 2015-06-16
Remove wordwrap which can break parsing lines in treeherder

gregarndt:


gregarndt:
can you give this a quick sanity check?  This is so that we don't do something like "taskcluster Error: something bad happened. Reason: \n worker-shutdown" where worker-shutdown is on a separate line and doesn't get picked up by the TH log parser.  wordwrap was hopes to make things nice, but it only applied to docker-worker specific messages injected into the logs and everything else can be whatever line length.

jonasfj:
I'm okay with this it looks to work...

But wouldn't it be simpler to make `fmtErr` that does the same as `fmtLog` except it prefixes with `[taskcluster] Error:` for all lines... or something like that... Or would that make sheriffs sad as they would get multiple error lines.

jonasfj:
Maybe the pattern that TH looks for should have been `[taskcluster-error]`

jonasfj:
Anyways, I'm okay with you merge this.. but some of the messages are not very well suited for this...

PR #107, opened 2015-06-16
Docker exec - Interactive **ready for review**

jonasfj:
Just notes for development... Not a formal review yet.


lib/features.js
```diff
@@ -88,6 +88,14 @@ const features = {
                  'capabilities or host volume mounts.',
     defaults: false,
     module: require('./features/dind')
+  },
+
+  websocketServer: {
```
jonasfj: This is the name that will be referenced in:
```js
task.payload.features = {
  websocketServer: true
}
```
I think this is too generic... Let's make it `interactive` that's more intuitive.. .

lib/features.js
```diff
@@ -88,6 +88,14 @@ const features = {
...
+    title: 'Docker Exec Websocket Server',
+    description: 'Runs another process alongside the original executed process' + 
+                 'and pipes the result through websockets.',
```
jonasfj: These docs should probably focus on how you can use this to fake "SSH" into the container...

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,43 @@
...
+      port = Math.floor((Math.random() * (61000 - 32768)) + 32768); 
+      try {
+        await new Promise((resolve, reject) => {
+          httpServ.listen(port);
+          httpServ.once('listening', () => {
+            httpServ.removeListener('error', reject);
+            resolve();
+          });
+          httpServ.once('error', reject)};
+      }
```
jonasfj: Missing `catch` clause...

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  async created (task) {
+    debug('creating ws server');
+    //TODO: add https support
+    var httpServ = http.createServer();
+    var port;
+    //searching for an open port between 32768 and 61000
+    for (let i = 0; i < 100; i++)
+    {
```
jonasfj: nit, but line break is here is a bit excessive...
Anyways, I wouldnt use a for-loop they are bad because once you pass `100` attempts it just continues without throwing an error... That's not good...

I would do:
```js
var attempts = 0;
while (true) {
  port = ...
  try {
    ....
  } catch (err) {
    if (err.code !== 'EADDRINUSE') {
      throw err;
    }
    attempts += 1;
    if (attempts >= MAX_RANDOM_PORT_ATTEMPTS) {
      throw err;
    }
  }
}
```

Then at the top of the file declare:
```js
const MAX_RANDOM_PORT_ATTEMPTS = 20;
```

Btw, love that you remove event handler for `error` on success...

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,42 @@
...
+    this.path = slugid.v4();
+  }
+
+  async created (task) {
+    debug('creating ws server');
+    //TODO: add https support
+    var httpServ = http.createServer();
+    var port;
+    //searching for an open port between 32768 and 61000
+    for (var i = 0; i < 10; i++) {
```
jonasfj: I wouldn't use a for-loop they are bad because once you pass `100` attempts it just continues without throwing an error indicating that we failed to start listening... That's not good... Maybe it's a critical error, maybe not, but whatever it is it should be logged :)

I would do:
```js
var attempts = 0;
while (true) {
  port = ...
  try {
    ....
  } catch (err) {
    // Only handle address in use errors.
    if (err.code !== 'EADDRINUSE') {
      throw err;
    }
    attempts += 1;
    if (attempts >= MAX_RANDOM_PORT_ATTEMPTS) {
      throw err;
    }
  }
}
```

Then at the top of the file declare:
```js
const MAX_RANDOM_PORT_ATTEMPTS = 20;
```

Btw, love that you remove event handler for `error` on success...

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,42 @@
...
+          resolve();
+        });
+        httpServ.once('error', reject)})).then(() => {i = 100;});
+    }
+
+    this.server = new DockerExecServer({
+      server: httpServ,
+      containerId: task.dockerProcess.id,
+      path: this.path,
+    });
```
jonasfj: I might be wrong... but aren't you supposed to create this before listening? I'm not sure it matter though :)

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,42 @@
...
+    //TODO: create artifact via taskcluster-client here later 
+  }
+
+  async killed (task) {
+    this.server.close();
```
jonasfj: You should `await` here... Or you can make the function non-async and `return this.server.close()`...

**Update** I just realized the function is synchronous...`server.close()`, but we probably should cleanup correctly someday... and return errors if cleanup fails.

package.json
```diff
@@ -10,7 +10,7 @@
     "test": "test"
   },
   "scripts": {
-    "test": "./test/test.sh $(find test -name '*_test.js') --reporter spec"
+    "test": "./test/test.sh"
```
jonasfj: Really?

Jeg run it with commandline instead of modifying the `package.json` unless this make sense in general...

test/integration/docker_exec_websocket_server_test.js
```diff
@@ -0,0 +1,41 @@
...
+        command: cmd('sleep 30'),
+        maxRunTime: 40,
+        features: {
+          websocketServer: true
+        }
+      }
+    };
+
+    let result = await worker.postToQueue(task);
+    // var client = new DockerExecClient({
```
jonasfj: The trick here is going to be: 
```js
var resultPromise = worker.postToQueue(task);
var client = new DockerExecClient({....
.... // play with client

await resultPromise;
```

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+    debug('port found', port);
+    this.server = new DockerExecServer({
+      server: httpServ,
+      containerId: task.dockerProcess.id,
+      path: this.path,
+    });
+    this.url = url.format({
+      protocol: 'ws',
```
jonasfj: We need to make this `wss` when we add SSL support, we can probably use a self-signed cert for localhost when testing... We just disable cert validation I have code for this...

I know DockerExecServer doesn't have ssl support yet... I'm just saying this is likely next step.

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,80 @@
...
+      slashes: true,
+      hostname: task.hostname,
+      port: port,
+      pathname: this.path,
+    });
+    debug(this.url);
+
+    var queue = task.runtime.queue;
+    var expiration = new Date(Date.now() + 60 * 60 * 24);
```
jonasfj: How about using `maxRuntime` from `task.payload`... Checkout what `task.js` does, I'm pretty we have a max run time limit which is enforced somewhere... It seems like that would be a good expiration time :)

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,80 @@
...
+    await queue.createArtifact(
+      task.status.taskId,
+      task.runId,
+      'interactive', {
```
jonasfj: A better name would be `private/mozilla/interactive.sock`, but long term this needs to be configurable.

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,80 @@
...
+        storageType: 'reference',
+        expires: expiration.toJSON(),
+        contentType: 'text/plain', //any way to fix this?
```
jonasfj: `application/octet-stream` is probably the right thing for binary thing we don't know what is :)

lib/features/docker_exec_websocket_server.js
jonasfj: Note, I would like to add one additional artifact named `private/mozilla/interactive` that links to:
`https://tools.taskcluster.net/interactive/?taskId=<taskId>&runId=<runId>&socket=<this.url>`.

test/integration/docker_exec_websocket_server_test.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+      }
+    };
+    debug('posting to queue');
+
+    var client;
+
+    let resultPromise = worker.postToQueue(task, taskId);
+    var passed = false;
+    setTimeout(async () => {
```
jonasfj: Instead of doing `setTimeout` you have two better options:
 1) `await base.testing.sleep(2000)` (basically a sleep utility)
 2) Use `await base.testing.poll(async () => { /* something that can fail */ })`

(2) is the most reliable one... It'll simply retry until the function given succeeds. Note that it does fail after some time and number of retries, this is configurable, see:
https://github.com/taskcluster/taskcluster-base/blob/master/testing.js#L529-L539

By default it'll try 16 times, and sleep 250ms between each retry... This is more robust than sleeping and then option docker boots quickly. Imagine a case where docker has to pull the image or something.

test/integration/docker_exec_websocket_server_test.js
```diff
@@ -0,0 +1,92 @@
...
+    if (worker) {
+      await worker.terminate();
+      worker = null;
+    }
+  });
+  test('cat', async () => {
+    let taskId = slugid.v4();
+  	let task = {
+      payload: {
+        image: 'taskcluster/test-ubuntu',
```
jonasfj: I think you'll like the `busybox` image more, it's really just 5mb :) and can sleep too...

test/integration/docker_exec_websocket_server_test.js
```diff
@@ -0,0 +1,92 @@
...
+      assert(url, 'artifact not found');
+
+      client = new DockerExecClient({
+        tty: false,
+        command: ['cat'],
+        url: url,
+      });
+      await client.execute();
+
+      var buf = new Buffer([0xfa, 0xff, 0x0a]);
```
jonasfj: I'm curious, why not test with something like: `crypto.randomBytes`
See: https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

If we generate say 2mb or 20mb of random data that would prove that it can handle binary data.

Oh, actually it doesn't prove that it's interpreted correctly inside the container... Just that it comes in and out as the same... Which IMO is super close, but maybe it would be interesting to run `md5sum` inside the container and then compare the checksum written out. But that's probably a different test case.

test/integration/docker_exec_websocket_server_test.js
```diff
@@ -0,0 +1,92 @@
...
+    }, 20000);
+    var promise = new Promise((resolve, reject) => {
+      setTimeout(() => {
+        if (passed) {
+          resolve();
+        } else {
+          reject();
+        };
+      }, 30000);
+    });
```
jonasfj: Just `await new Promise(accept => client.stdout.once('end', accept));` and then `assert(passed)`...

Also doesn't the client have an event for exitcode? We should check that too...

Sleeps like this are bad in tests... they make test unreliable and they make tests slow.. .It's best to wait for an event, or if not possible use `base.testing.poll`. Note, normally we can't wait on an event like I just did above, not unless we are 100% sure the event is going to arrive as it could block the promise if there is an error and the `end` event never happens. I think (not sure) the end event always happens.. eitherway for testing it's acceptable that a bug manifests itself in a timeout error. It's not idea, but also not worth investing huge effort in fixing :)

jonasfj:
@EggyLv999, as you said it needs clean up, but it looks rather good already...
We have some extra features to add and more tests to write, but it looks very promising :)

lib/features/docker_exec_websocket_server.js
jonasfj: I was just talking to garndt, and the ssl certificate used for testing:
https://github.com/taskcluster/docker-worker/blob/docker-exec/config/test.js#L22-L25

Is signed for `localhost`, it's self-signed so not really worth much and won't validate. But if we set `rejectUnauthorized: false,` when we make the request node won't reject an unauthorized certificate. This is only something we should do in testing, but it's easily possible see:
https://github.com/websockets/ws/blob/master/doc/ws.md#new-wswebsocketaddress-protocols-options

We just need to allow the option in `DockerExecClient` and pass it on to `ws`. Note, it's important that `rejectUnauthorized` is `true` by default, ether `true` or not specified as a property on the options object given to `ws`. I'm not sure `rejectUnauthorized: undefined` is good enough.

------

Note, the ssl certificate filenames are stored in `task.runtime.ssl.certificate` and `task.runtime.ssl.key`, see: 
https://github.com/taskcluster/docker-worker/blob/docker-exec/lib/features/local_live_log.js#L75-L76
Note it seems we don't even need to modify `DockerExecServer` here, we just use `require('https')` and give it there, see:
https://github.com/websockets/ws/blob/master/examples/ssl.js#L34-L35

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,80 @@
+const MAX_RANDOM_PORT_ATTEMPTS = 20;
+
+var debug = require('debug')('docker-worker:features:docker_exec_websocket_server.js');
+var DockerExecServer = require('docker-exec-websocket-server').DockerExecServer;
+// var waitForPort = require('../wait_for_port');
+var http = require('http');
```
jonasfj: Let's use `https` instead...

lib/features/docker_exec_websocket_server.js
```diff
@@ -0,0 +1,80 @@
...
+export default class WebsocketServer {
+  constructor () {
+    this.path = '/' + slugid.v4();
+    // this.path = '/a';
+  }
+
+  async started (task) {
+    debug('creating ws server');
+    //TODO: add https support
+    var httpServ = http.createServer();
```
jonasfj: Just do:
```js
var httpServ = https.createServer({
  key: fs.readFileSync(task.runtime.ssl.key),
  cert: fs.readFileSync(task.runtime.ssl.certificate)
});
```
And we should have HTTPS.

Note, we might want to read the files asynchronously later... using something like:
```js
var readFile = Promise.denodeify(fs.readFile);
var [key, cert] = await Promise.all([
  readFile(task.runtime.ssl.key),
  readFile(task.runtime.ssl.certificate)
]);
var httpServ = https.createServer({key, cert});
```
I live the unpacking and packing in ES6 so nice :)

lib/features.js
```diff
@@ -88,6 +88,15 @@ const features = {
                  'capabilities or host volume mounts.',
     defaults: false,
     module: require('./features/dind')
+  },
+
+  interactive: {
+    title: 'Docker Exec Websocket Server',
+    description: 'Runs another process alongside the original executed process' + 
+                 'and pipes the result through websockets. Can be used for SSH-like' +
+                 'access to docker containers.',
```
jonasfj: Nit, but this is the description that a task writer sees.

So probably has to be:
This allows you to interactively run commands inside the container and attaches you to the stdin/stdout/stderr over a websocket. Can be used to for SSH-like...

lib/features/interactive.js
```diff
@@ -0,0 +1,93 @@
+const MAX_RANDOM_PORT_ATTEMPTS = 20;
```
jonasfj: Nitm
I would put this after `require` statements...
And give a doc comment...

lib/features/interactive.js
```diff
@@ -0,0 +1,93 @@
...
+var fs = require('fs');
+
+export default class WebsocketServer {
+  constructor () {
+    this.path = '/' + slugid.v4();
+  }
+
+  async started (task) {
+    debug('creating ws server');
+    var httpsOpt = true; //set this somewhere else later
```
jonasfj: I might be wrong here but I think all config options are added to ´task.runtime`. So we can just add a config option in `config/defaults.js` called `interactive: {ssl: true}` and then get it here using `task.runtime.interactive.ssl`.

This is likely a good option for many config options that we configure on a per worker level.

lib/features/interactive.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+    //and its corresponding url
+    this.url = url.format({
+      protocol: httpsOpt ? 'wss' : 'ws',
+      slashes: true,
+      hostname: task.hostname,
+      port: port,
+      pathname: this.path,
+    });
+    debug(this.url);
```
jonasfj: Is this necessary... I'm thinking we log this, and this url is a bit sensitive...

lib/features/interactive.js
```diff
@@ -0,0 +1,93 @@
...
+
+    var queue = task.runtime.queue;
+    await queue.createArtifact(
+      task.status.taskId,
+      task.runId,
+      'private/mozilla/interactive.sock', {
+        storageType: 'reference',
+        expires: expiration.toJSON(),
+        contentType: 'application/octet-stream',
+        url: this.url
```
jonasfj: nit, but there is no reason to store url on `this`...
However, it would be best to name it something other than url... as it otherwise overwrite the `url` module you've required.

I suspect `socketUrl` or something would do...

jonasfj: Also let's add the other artifact we talked about... even if the `https://tools.taskcluster.net/interactive/?socket=,runId=,taskId=` link won't work yet... it will soon, and we don't want to update docker-worker when it does :)

test/integration/interactive_test.js
```diff
@@ -0,0 +1,149 @@
...
+      url = await getWithoutRedirect(signedUrl);
+      assert(url, 'artifact not found');
+    }, 20, 1000);
+
+    //for testing, we don't care about https verification
+    var client = new DockerExecClient({
+      tty: false,
+      command: ['cat'],
+      url: url,
+      wsopts: {rejectUnauthorized: false}
```
jonasfj: neat, hack I like that... people will rarely need to provide options for `ws` anyways :)

test/integration/interactive_test.js
```diff
@@ -0,0 +1,149 @@
...
+    });
+    await client.execute();
+
+    const TEST_BUF_SIZE = 1024 * 1024;
+
+    var buf = await Promise.denodeify(crypto.pseudoRandomBytes)(TEST_BUF_SIZE);
+    // make sure the last character is a newline
+    buf[TEST_BUF_SIZE - 1] = 0x0a;
```
jonasfj: curious, is that really necessary? Won't cat print if you signal end to input...
Oh, we probably need to have a message for signaling `end` of input... 

Basically if someone does `client.stdin.end()` they're saying that this is the end of the stream and it should close. Do we even send that right now?

If not then isn't the `cat` process killed by the docker container coming to an end.. .rather than terminating because the input stream is at an end... So we don't get exit code zero... And the cat subprocess remains alive forever.

Note, to close bash/cat/grep we just end stdin... This is how `cat myfile | grep needle` works... grep terminates because the stream given by stdin comes to an end. We need the ability to close the stream, note that this is different from aborting or killing the process... Because the process could still be writing back to us for a while before it stops... or it could choose not to stop or even care about the stdin...

To do real kill we probably need some tricks and smart utilities, but let's do that later...

lib/features/interactive.js
jonasfj: Actually, let's make it `https://tools.taskcluster.net/interactive/?socket=...&runId=...&taskId=...&type=docker-exec-ws-v1`

lib/features/interactive.js
```diff
@@ -0,0 +1,112 @@
+var debug = require('debug')('docker-worker:features:interactive');
```
gregarndt: this is a nit, but since we're adopting es6/7 goodies elsewhere, I'm not sure why we're not using import and let

lib/features/interactive.js
```diff
@@ -0,0 +1,112 @@
...
+export default class WebsocketServer {
+  constructor () {
+    this.path = '/' + slugid.v4();
+  }
+
+  async started (task) {
+    debug('creating ws server');
+    var httpServ;
+    if (task.runtime.interactive.ssl) {
+      var readFile = Promise.denodeify(fs.readFile);
```
gregarndt: we use 'mz' package elsewhere so it might be easy to just use that here instead of promisefying it.

lib/features/interactive.js
```diff
@@ -0,0 +1,112 @@
...
+      var [key, cert] = await Promise.all([
+        readFile(task.runtime.ssl.key),
+        readFile(task.runtime.ssl.certificate)
+        ]);
+      httpServ = https.createServer({key, cert});
+    } else {
+      http.createServer();
```
gregarndt: should this be httpServ = http.createServer() ?

test/integration/interactive_test.js
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  async function getWithoutRedirect (url) {
+    var res = await request.get(url).redirects(0).end();
+    return res.headers.location;
+  };
+
+  test('cat', async () => {
+    let taskId = slugid.v4();
+  	let task = {
```
gregarndt: nit: indentation should be reduced here

test/integration/interactive_test.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+  });
+
+  async function getWithoutRedirect (url) {
+    var res = await request.get(url).redirects(0).end();
+    return res.headers.location;
+  };
+
+  test('cat', async () => {
+    let taskId = slugid.v4();
```
gregarndt: not sure between the mixture between let and var here

lib/task.js
```diff
@@ -665,6 +668,24 @@ export default class Task {
...
     this.runtime.log('task run');
+
+    // Call started hook when container is started
+    dockerProc.once('container start', async (container) => {
+      try {
+        await this.states.started(this);
+      }
+      catch (e) {
+        debug(e.stack);
+        return await this.abortRun(
```
gregarndt: Do we have a test verifying that the run is aborted and marked as failure if the right conditions are present?  Could do this by changing the startup configuration of the worker to have an invalid certificate, or move the port range to the configuration and then you can specify an invalid port range (or a known used port) as part fo the startup config to make sure it fails gracefully.

jonasfj: Can't we just make a special "feature" that we only load during testing?

gregarndt: Sure, we could, which would mean that we have mock/test features in features.js right?  Unless we offload that to somewhere else.  Also, that only tests the logic here of the 'started' state handling.

Although if the purpose of just testing the logic specifically there (the try/catch), then it would easily be tested just by doing what we do in other tests and specify an invalid key path which would test to make sure not only that the feature throws an exception when the key can't be found, but also that we handle it here properly.

EggyLv999:
Just leaving a note here: for this to work, the version of node inside the docker container must match the version of node on which npm install was called. 10 and 11 are compatible, but 10/12 and 11/12 will not work due to ws being a binary module and the binary api for node changing between versions 11 and 12.

EggyLv999:
This PR has been obsoleted by #121.

PR #109, opened 2015-06-19
Bug 1175315 - Add Proxy feature for the releng API

gregarndt:
This makes use of the relengapi proxy that Dustin created.  It was successfully used here:
https://tools.taskcluster.net/task-inspector/#ELmePgAbRz-W_7A_RUWmSw/0

This is probably the bare minimum stuff that's needed to get a feature like this landed.  Very basic.

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -41,6 +41,13 @@ module.exports = {
     randomizationFactor: 0.25
   },
 
+  features: {
+    relengAPIProxy: {
+      image: 'quay.io/djmitche/relengapi-proxy:0.0.2',
+      token: process.env.RELENG_API_TOKEN
+    }
+  },
+
   ssl: {
     certificate: '/etc/star_taskcluster-worker_net.crt',
     key: '/etc/star_taskcluster-worker_net.key'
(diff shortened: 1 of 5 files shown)


config/defaults.js
```diff
@@ -41,6 +41,13 @@ module.exports = {
     randomizationFactor: 0.25
   },
 
+  features: {
```
gregarndt: note here: I would like to move towards grouping feature related config together.  in a future PR if this method is acceptable, I can move the images we store in the config under features:<name>:image:<image>

config/defaults.js
```diff
@@ -41,6 +41,13 @@ module.exports = {
...
+    relengAPIProxy: {
+      image: 'quay.io/djmitche/relengapi-proxy:0.0.2',
+      token: process.env.RELENG_API_TOKEN
```
gregarndt: This is only necessary right now until secrets land and we can stuff it in there.

lib/features/releng_api_proxy.js
```diff
@@ -0,0 +1,75 @@
...
+      Tty: true,
+      AttachStdin: false,
+      AttachStdout: true,
+      AttachStderr: true,
+
+      Cmd: cmd
+    });
+
+    // Terrible hack to get container promise proxy.
+    this.container = docker.getContainer(this.container.id);
```
jonasfj: nit, upgrade `dockerode-promise` and this _might_ not be needed anymore...

jonasfj:
Looks good to me :)

PR #110, opened 2015-06-29
Update vagrant and docker-worker-test to node 0.12

gregarndt:


lib/features/artifacts.js
```diff
@@ -128,35 +129,42 @@ export default class Artifacts {
...
+      var putReq = https.request(options);
 
       // Stream tar entry to request before ending the request
       stream.pipe(putReq);
-      putReq.end();
 
-      // Wait until the response is sent.
-      var res = yield waitForEvent(putReq, 'response');
+      let response = yield new Promise((accept) => {
+        putReq.on('response', (res) => { accept(res); });
```
jonasfj: What about errors? `putReq.on('error', reject);`

Note, I would fully support a method like `uploadS3Artifact(stream, putUrl, contentType, contentLength)` that returns a promise... So we can use between different files..

test/integration/logging/local_live_log_test.js
```diff
@@ -67,8 +67,10 @@ suite('live logging', () => {
     let artifactUrl = `https:\/\/queue.taskcluster.net/v1/task/${taskId}/runs/0/artifacts/public/logs/live.log`
 
     // Don't follow redirect, we just care about where it's going
-    let res = await request.get(artifactUrl).redirects(0).end();
```
jonasfj: Pretty sure `redirect(0)` still works... It's just that it now throw an error on non-2xx reponse codes..

jonasfj: Problem with this approach is if we don't redirect to a valid URL...

jonasfj:
Handle errors from requests... and try to deduplicate the logic. Otherwise, this looks reasonably good.

jonasfj:
I don't see a way to easily implement retries... So I'm okay with us leaving that for later... though it's really sad that we don't have retries...

But if the upload stream logic is in a  separate file then we can add it later.. not easy for easier :)

PR #115, opened 2015-07-07
Support retrieving secrets and userdata from provisioner

gregarndt:
The main purpose of this is to consume user-data from the data object within user-data .  Not consuming this correctly broke a few workers where we were specifying instance type user-data overrides.   This should also all moving secrets into the provisioner instead of baking in the AMIs


gregarndt:
Care to give this a review?  Once we land this, I can work on moving credentials to the worker definitions.

gregarndt:
Also, the worker type running CI tests is already using an AMI based on these changes because it's the only way to get CI tests to run.

lib/host/aws.js
```diff
@@ -83,32 +95,40 @@ export async function configure(baseUrl) {
...
-      minimumCycleSeconds: minutes(2),
+      minimumCycleSeconds: minutes(2)
     }
   };
 
-  log('metadata', config);
+  let userdata = await getJsonData(`${baseUrl}/user-data`);
+  let userSuppliedData = userdata.data;
+  let securityToken = userdata.securityToken;
+  let provisionerBaseUrl = userdata.provisionerBaseUrl;
```
jonasfj: This is probably not more readable; i'm just showing off my skillz :)
```js
let {data, securityToken, provisionerBaseUrl} = userdata;
```
Also works....

lib/host/aws.js
```diff
@@ -83,32 +95,40 @@ export async function configure(baseUrl) {
...
+    taskcluster: secrets.credentials
+  };
+  log('read secrets');
+
+  await provisioner.removeSecret(securityToken);
+
+  // Log config for record of configuration but without secrets
+  log('config', config);
+
+  return Object.assign(config, userdata, userSuppliedData, credentials, secrets.data);
```
jonasfj: Isn't it wrong to have `userdata` here... You only have it here for legacy right?

Do we need it, if not... let's remove it, then we don't have to have the `delete userdata.data` statements either.

jonasfj:
Looks fairly good to me `r+`, but please do review if `userdata` should be used in `Object.assign`.

lib/host/aws.js
gregarndt: Well really the main thing we care about in userdata is capacity, but instead of just pulling out one value, I just figured it was easier to remove the couple of things we know we don't need and we accept the rest.  This way if there are ever other values that the provisioner supplies, we have it in the config.

lib/host/aws.js
gregarndt: ah, I keep forgetting about the dereferencing (?) we can do.

lib/host/aws.js
jonasfj: > This way if there are ever other values that the provisioner supplies, we have it in the config.

@gregarndt, I think this is a bad idea... IMO config shouldn't be dynamically typed. A newer provisioner version could add value that breaks old versions of docker-worker. It's better to be explicit about it.

Note, I try to also be explicit about reading from config, by using `config.get()`. I notice that docker-worker isn't doing this, but I suspect being explicit about reading configs is a bit overkill... However, writing config, is probably best to be explicit about what goes in where from and from where..

In this case just do: `Object.assign(config, {capacity: userdata.capacity}, userdata.data, {taskcluster: secrets.credentials}, secrets.data);` I think it's much safer...

gregarndt: I definitely like this idea and I'll make the changes.  Awesome suggestion, thank you for making me see the errors of my ways :)

PR #121, opened 2015-07-08
Docker exec flock

EggyLv999:
And here is the long-awaited PR for the interactive feature. I have no idea how a couple of random 1-line fixes are showing up here, but they are.


jonasfj:
@EggyLv999, tests doesn't seem to pass.. The `AssertionError: auth download works` error is known and @gregarndt  is working on it. So that doesn't block this PR, but `AssertionError: artifact not found` does block this PR, especially since I think it's your tests :)

Vagrantfile
```diff
@@ -1,6 +1,6 @@
 Vagrant.configure("2") do |config|
   config.vm.box = "phusion/ubuntu-14.04-amd64"
-
+
```
jonasfj: Nit, but I think this change can be removed :)

lib/features/interactive.js
```diff
@@ -0,0 +1,151 @@
...
+    this.lockFile = await Promise.denodeify(fs.open)(this.lock, 'w');
+
+    return {
+      binds: [{
+        source: path.join(__dirname, '../../bin-utils'),
+        target: '/.taskcluster_utils',
+        readOnly: true
+      }, {
+        source: this.lock,
+        target: '/tmp/interactive.lock',
```
jonasfj: Thinking about this... this could by freak accident cause a collision with something inside the container.
Or `/tmp` could not exist... I'm thinking that calling it `/.taskcluster-interactive-session.lock` is safer. It's a small change anyways.

lib/features/interactive.js
```diff
@@ -0,0 +1,151 @@
...
+    this.lock = path.join('/tmp/', id + '.lock');
+  }
+
+  async link(task) {
+    this.lockFile = await Promise.denodeify(fs.open)(this.lock, 'w');
+
+    return {
+      binds: [{
+        source: path.join(__dirname, '../../bin-utils'),
+        target: '/.taskcluster_utils',
```
jonasfj: Hmm... Actually looking at docker, I see that it already has a `/.dockerenv` and `/.dockerinit` files.
How about we follow that convention and make this `/.taskclusterutils/`. Then we also avoid any complications with dash and underscore...

lib/features/interactive.js
```diff
@@ -0,0 +1,151 @@
...
+    while (true) {
+      port = Math.floor((Math.random() * (61000 - 32768)) + 32768); 
+      try {
+        await (new Promise((resolve, reject) => {
+          httpServ.listen(port);
+          httpServ.once('listening', () => {
+            httpServ.removeListener('error', reject);
+            resolve();
+          });
+          httpServ.once('error', reject)}));
```
jonasfj: if we more than `12` iterations or so node will complain about event handler leaking... It's not really a problem, because promise won't resolve more than once. But if we don't remove the `error` handler  and we retry to we change the semantics... Not handling an event called `error` is a special case that crashes the process.

```js
await new Promise((resolve, reject) => {
  let listeningHandler, errorHandler;
  listeningHandler = () => {
    httpServ.removeListener('error', errorHandler);
    resolve();
  };
  errorHandler = (err) => {
    httpServ.removeListener('error', listeningHandler);
    reject(err);
  };
  httpServ.once('listening', listeningHandler);
  httpServ.once('error', listeningHandler);
  httpServ.listen(port);
});
```

Note, we might also want to listen for the `error` event from `httpServ` or not... I'm not sure what we would do if it happens... I suspect it doesn't really happen once we've started listening anyways.

config/defaults.js
```diff
@@ -184,5 +184,13 @@ module.exports = {
...
 
-  dockerWorkerPrivateKey: '/etc/docker-worker-priv.pem'
+  dockerWorkerPrivateKey: '/etc/docker-worker-priv.pem',
+
+  interactive: {
+    ssl: true,
+    //time, in seconds, between end of task and end of interactive session
+    //could be shorter due to worker node shutdown
+    minTime: 10 * 60,
+    expirationAfterSession: 5 * 60
```
jonasfj: Comment is wrong...

`minTime` is minimum time the container will exist.. So the minimum time you have to connect. But also time that will definitely be wasted, if people run a very short task... We might want to keep this lower, say 3 min, if people submit tasks with interactive enabled by default having it at 10min could cost a lot of waste time, right?

`expirationAfterSession` is time wasted after an interactive session... I have no problem with that being high... We rarely have interactive sessions...

lib/features/interactive.js
```diff
@@ -0,0 +1,151 @@
...
+
+    let socketArtifact = queue.createArtifact(
+      task.status.taskId,
+      task.runId,
+      'private/mozilla/interactive.sock', {
+        storageType: 'reference',
+        expires: expiration.toJSON(),
+        contentType: 'application/octet-stream',
+        url: socketUrl
+      });
```
jonasfj: Intentation... I think you'll want to take it two space away... nit..

lib/features/interactive.js
```diff
@@ -0,0 +1,151 @@
...
+        contentType: 'text/html',
+        url: url.format({
+          protocol: 'https',
+          host: 'tools.taskcluster.net',
+          pathname: '/interactive',
+          query: {
+            taskId: task.status.taskId,
+            runId: task.runId,
+            socket: socketUrl
+          }
```
jonasfj: Just a suggestion... but how about we add a third parameter: `v: '1'` to the querystring...
So we can make version 2, in some future if we ever have to, I doubt it, the protocol seems solid to me.

lib/features/interactive.js
```diff
@@ -0,0 +1,151 @@
...
+
+    await Promise.all([socketArtifact, toolsArtifact]);
+    debug('artifacts made');
+  }
+
+  async killed (task) {
+    if(this.server) {
+      this.server.close();
+    }
+    fs.unlink(this.lock);
```
jonasfj: Let's handle any errors from unlink, at least log them... Maybe with `"[alert-operator]"` in the debug message... I think we can agree that something is very wrong if the file was deleted...

Note, if the file is deleted inside the container, it'll just be unlinked, it won't be unlinked outside the container (or so I think).

lib/shared_file_lock.js
```diff
@@ -0,0 +1,48 @@
+import assert from 'assert';
+import Debug from 'debug';
+import fs from 'fs-ext';
+import Promise from 'promise';
+
+let debug = Debug('docker-worker:lib:shared_file_lock');
+
+export default class SharedFileLock {
+  constructor(lockFile) {
```
jonasfj: Let's add a comment to `SharedFileLock` saying what it does... And one to the constructor saying that it takes a file handle, right? (not sure)

lib/shared_file_lock.js
```diff
@@ -0,0 +1,48 @@
...
+    this.count += 1;
+    debug('acquire; count is %s', this.count);
+  }
+
+  //releases a lock after some delay, at 0 locks it will unlock the lockfile
+  async release(delay = 0) {
+    if(delay > 0) {
+      return(setTimeout(() => {this.release()}, delay));
+    }
+    assert(this.count > 0, "Internal error");
```
jonasfj: Probably need to log with the string `"[alert-operator]"`.

lib/shared_file_lock.js
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+    this.count += 1;
+    debug('acquire; count is %s', this.count);
+  }
+
+  //releases a lock after some delay, at 0 locks it will unlock the lockfile
+  async release(delay = 0) {
+    if(delay > 0) {
+      return(setTimeout(() => {this.release()}, delay));
```
jonasfj: nit, but you can do with: `return setTimeout(() => this.release(), delay);`

lib/shared_file_lock.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+    assert(this.count > 0, "Internal error");
+    this.count -= 1;
+    if(this.count === 0 && this.locked) {
+      let err = await Promise.denodeify(fs.flock)(this.lockFile, 'un');
+      if(err) {
+        debug('couldn\'t acquire lock, this is probably bad');
```
jonasfj: This we probably have to lock with `[alert-operator]` because it's very very bad... Hmm.. may be bad, I'm not sure..

Hmm... I just realized we might not have released all locks when the task is killed, if it's killed by say timeout of `MaxRuntime`...

lib/features/interactive.js
jonasfj: Actually, why not add this to `SharedFileLock`, the it can just ignore calls to `release` after the file has been deleted.

Well, it can release the lock if held, before it deletes the file and then delete the file... Note, that anyone else having a file handle can still hold a lock on the file or try to lock it... so unlocking before we unlink might be nice.

I think this mainly has effect if we someday use this lock in more than one feature or something. But either way it's nice to keep things clean..

lib/shared_file_lock.js
EggyLv999: I'm fairly certain that a dead process can't hold locks, so I don't think we have to worry about that.

jonasfj: Sorry, didn't mean release all locks... I meant we might not have called `semaphore.release()` yet.. So count isn't zero.

lib/task.js
```diff
@@ -260,6 +263,21 @@ export default class Task {
       procConfig.create.HostConfig.Binds = binds;
     }
 
+    if(this.task.payload.features.interactive) {
+      //TODO: test with things that aren't undefined
```
jonasfj: Hmm.. yeah, let's add a test case for this...

lib/task.js
```diff
@@ -260,6 +263,21 @@ export default class Task {
...
+      let oldEntrypoint = (await this.runtime.docker.getImage(config.image).inspect()).Entrypoint;
+      if(typeof oldEntrypoint === 'string') {
+        oldEntrypoint = ['/bin/sh', '-c', oldEntrypoint];
+      } else if(typeof oldEntrypoint === 'undefined') {
```
jonasfj: nit, but `oldEntrypoint === undefined` works too...

lib/shared_file_lock.js
```diff
@@ -0,0 +1,70 @@
...
+  constructor(lockFile) {
+    this.count = 0;
+    this.lockFile = lockFile;
+    this.locked = false;
+  }
+
+  //acquires a lock, at >=1 locks it will flock the lockfile
+  async acquire() {
+    if(!this.lockFd) {
+      this.lockFd = await mzfs.open(this.lockFile, 'r');
```
jonasfj: I actually liked that you were giving it a file handle... because even if the file is unlinked (deleted) the file handle will still exist, so you should always be able to at least attempt to lock and unlock...

Now we have another possible error case where container and session could race to delete the file... cause this to throw and exception... well, okay I'm not sure deleting the file inside the container deletes it outside. But locking the file handle that was used to create the file was quiet brilliant design in the first place...

I just wanted you to document, that the constructor took a file handle  and not a file name.. 

Note, another option is to move `await mzfs.open(this.lockFile, 'r');` into a setup method, and call that setup method in place of `await fs.writeFile(this.lock);` in `lib/features/interactive.js`..

I think this is mostly nit picking... but with locks it's important to be careful and not allow "interesting" things to happen. Too often it leads to hard to debug issues and deadlocks. Particularly, if someday we were to make another plugin/feature that wanted to lock.

deploy/packer/app/scripts/deploy.sh
```diff
@@ -33,3 +33,7 @@ sudo npm install -g babel@4.7.16
 sudo sh -c 'echo "v4l2loopback" >> /etc/modules'
 sudo sh -c 'echo "snd-aloop" >> /etc/modules'
 sudo depmod
+
+# install busybox
+sudo curl -o /worker/bin-utils/busybox http://www.busybox.net/downloads/binaries/latest/busybox-x86_64
```
gregarndt: So, I see that this will be done on the ec2 instances as well as the changes made for the vagrant environment, but this file needs to be present in our ci image I would suspect

jonasfj: I suggested that we move into the `postinstall` hook for `npm install`... That way it'll always be there when we've done npm install... And it'll in in the ci environment...

gregarndt: Fantastic, thanks.

lib/features/interactive.js
```diff
@@ -0,0 +1,169 @@
...
+    } else {
+      httpServ = http.createServer();
+    }
+    debug('server made')
+
+    let port;
+    //searching for an open port between 32768 and 61000
+    let attempts = 0;
+    while (true) {
+      port = Math.floor((Math.random() * (61000 - 32768)) + 32768);
```
gregarndt: nit: trailing white space

test/integration/interactive_test.js
```diff
@@ -0,0 +1,332 @@
...
+import slugid from 'slugid';
+
+suite('use docker exec websocket server', () => {
+  let debug = Debug('docker-worker:test:interactive-test');
+
+  let worker;
+  // If taskcluster/artifact upload is under high load, this number needs to be adjusted up.
+  // It also causes the test to be slower by 2X that many seconds, so be careful with this.
+  // TODO: add polling to tests so they don't rely as much on this type of timing
+  let maxTime = 45;
```
gregarndt: nit: trailing white space

test/integration/interactive_test.js
```diff
@@ -0,0 +1,332 @@
...
+
+    await base.testing.sleep(maxTime * 1000 + 1000);
+    //should still be alive here, even though it was dead here last time
+    //This is because cat is still alive
+    let status = await worker.queue.status(taskId);
+    assert(status.status.state === 'running', 'stopped early!');
+    
+
+    client.close();
+    await base.testing.sleep(expTime * 1000 + 3000);
```
gregarndt: nit: trailing white space

EggyLv999:
Superceded by #134, a more recent version which taskcluster-ci loves more.

PR #122, opened 2015-07-09
Idle stats

EggyLv999:
Added the storage of some stats that let us know how efficient docker-worker really is.


lib/task_listener.js
```diff
@@ -229,7 +227,18 @@ export default class TaskListener extends EventEmitter {
     }
   }
 
-  addRunningTask(runningState) {
+  async recordCapacity () {
+    this.runtime.stats.record('capacityOverTime', {
+      duration: Date.now() - this.timestamp,
+      availableCapacity: await this.availableCapacity(),
```
jonasfj: I'm not sure this is good... `availableCapacity()` has side effects...

Perhaps we should just use the latest value returned from `availableCapacity()`?
@gregarndt, do you have any suggestions here...

jonasfj:
Apart from all the record keeping methods now being async and `availableCapacity()` having side effects that could be significant this looks good...

jonasfj:
@gregarndt, doesn't `availableCapacity()` have significant side effects on the device tester machines?

lib/task_listener.js
```diff
@@ -28,6 +25,7 @@ export default class TaskListener extends EventEmitter {
     this.runningTasks = [];
     this.taskQueue = new TaskQueue(this.runtime);
     this.taskPollInterval = this.runtime.taskQueue.pollInterval;
+    this.timestamp = Date.now();
```
gregarndt: nit: I probably would name this something more descriptive as to what you're trying to use it for (such as listenerStartTime)

lib/task_listener.js
gregarndt: Yea, we should probably memoize this for at least stats recording.  Most of the time this wouldn't be expensive to call, but because of the way testdroid is currently, this will cause requests to be made to bitbar to determine the available capacity.

gregarndt: could just set something like lastKnownCapacity each time availableCapacity is called, and the stats recording will just use lastKnownCapacity

jonasfj: `lastKnownCapacity` that sounds like a good idea...

lib/task_listener.js
```diff
@@ -229,7 +228,21 @@ export default class TaskListener extends EventEmitter {
     }
   }
 
-  addRunningTask(runningState) {
+  async recordCapacity () {
```
jonasfj: I don't think this or `addRunningTask` or `removeRunningTask` needs to be `async`...
If `lastKnownCapacity` hasn't been set can't we just use `0`. As far as I can see it'll never reach `recordCapacity` without first calling `getTasks` which calls `availableCapacity`...

So if we just set `this.lastKnownCapacity = 0` in the constructor I think we'll do fine. It's not even a bad assumption. If it's working on something and it never has checked for available capacity or attempted to poll for tasks then is it idle? Does it really have any available capacity?

I think it makes sense to initialize it to zero, it avoid an error that would **only** happen if some invariant was violated. Hence, an error that doesn't happen.

lib/task_listener.js
```diff
@@ -229,7 +228,21 @@ export default class TaskListener extends EventEmitter {
...
+    if(this.lastKnownCapacity === undefined) {
+      await this.availableCapacity();
+    }
+    this.runtime.stats.record('capacityOverTime', {
+      duration: Date.now() - this.lastTaskEvent,
+      availableCapacity: this.lastKnownCapacity,
+      runningTasks: this.runningTasks.length
```
jonasfj: Can we add one extra field called `totalCapacity: this.lastKnownCapacity + this.runningTasks.length`.
Because when `totalCapacity` is zero, we know the worker hasn't started yet... Or is waiting for a resource (say a device) before it can run anything.

Also it's nice to be able to say, what is average of `runningTasks` for all with `totalCapacity = 16` (ie big machines).
This would tell us if we utilize them. I know we can calculate this, but I suspect storing it would make it faster? Or, hmm maybe not... because it's not a tag... maybe we should add it as a string value so we can make a 0.9 tag.

lib/task_listener.js
```diff
@@ -253,6 +266,7 @@ export default class TaskListener extends EventEmitter {
       this.cleanupRunningState(runningState);
       return;
     }
+    await this.recordCapacity();
```
jonasfj: Please add a comment that it's important to call `recordCapacity` before we remove the entry from `this.runningTasks` because it counts it... Just a single line comment will do, so we don't move it by accident.

jonasfj: @EggyLv999 has a valid point that we're not calling `availableCapacity` when reporting a removed task... So we just removed what was there last time... that is not ideal...

Maybe we can add one to `lastKnownCapacity` when a task is removed...

lib/task_listener.js
```diff
@@ -229,7 +228,21 @@ export default class TaskListener extends EventEmitter {
...
   }
 
-  addRunningTask(runningState) {
+  async recordCapacity () {
+    if(this.lastKnownCapacity === undefined) {
+      await this.availableCapacity();
+    }
+    this.runtime.stats.record('capacityOverTime', {
+      duration: Date.now() - this.lastTaskEvent,
+      availableCapacity: this.lastKnownCapacity,
```
jonasfj: let's call it `idleCapacity`.

lib/task_listener.js
jonasfj: So add one to `lastKnownCapacity` after `recordCapacity`...

PR #123, opened 2015-07-10
Package cleanup2

EggyLv999:
This is like the less stable version of package_cleanup #118. I just updated a bunch of packages, hoping it doesn't break anything; going to let taskcluster sort this one out.


EggyLv999:
Not 100% sure that all of the bin changes work, they should be *theoretically* correct but you never know.

lib/worker_logs.js
```diff
@@ -1,6 +1,6 @@
-var fs = requrie('fs');
+var fs = require('fs');
```
EggyLv999: Given that require was spelled incorrectly here, is anything actually using this file? Can we just remove it? Nobody else has it required.

bin/worker.js
```diff
@@ -215,9 +212,9 @@ co(function *() {
   config.volumeCache = new VolumeCache(config);
 
   config.gc.on('gc:container:removed', function (container) {
-    container.caches.forEach(co(function* (cacheKey) {
-      yield config.volumeCache.release(cacheKey);
-    }));
+    container.caches.forEach(async (cacheKey) => {
```
gregarndt: Wouldn't we need to wait for all of these async promises to be completed?

lib/worker_logs.js
gregarndt: I'm not aware of what might be using this.  If this is not used anywhere I say remove it.

bin/worker.js
EggyLv999: The way I'm reading it, based on my limited knowledge of co, the original code doesn't wait for all of the promises. Not that it really matters, because none of this is really going to block because they're on a listener function. In fact, you could have it be something like 
```js
config.gc.on('gc:container:removed', () => {
  container.caches.forEach(config.volumeCache.release);
});
```
and it would perform identical to the original code, if that's what you want.

EggyLv999:
All this stuff is already merged in.

PR #124, opened 2015-07-10
Moved all co/yield to async/await, cleaned up packages a little

EggyLv999:
Changed the control flow of stuff. Everything here should be in working order.


EggyLv999:
#123 is a better version of this.

PR #127, opened 2015-07-14
Update worker configuration

gregarndt:
This is the start of work to remove secrets from the worker itself and move them into the provisioner.


gregarndt:
This is mainly switching to using lodash.defaultsDeep so that things like just the token could be specified for stateless hostname, but the other options are still taken from the original production config.

jonasfj:
Looks good to me, `r+` assuming you're already happy with it.

PR #132, opened 2015-07-16
Updated docker version and respawn issues

gregarndt:


lib/task_listener.js
```diff
@@ -77,6 +77,11 @@ export default class TaskListener extends EventEmitter {
   }
 
   async listenForCancelEvents() {
+    // Do not listen for events if pulse configuration is not supplied
```
gregarndt: This is mainly for when the worker for some reason restarts and can't retrieve the creds from secrets because they were removed.  Capacity is set to 0 and we expect the worker to die off, so it should not need to listen for cancel events anyways.

lib/stats/stat.js
```diff
@@ -12,7 +15,14 @@ export default class Stat {
     assert(config.workerNodeType, 'Worker instance type is required');
     assert(config.provisionerId, 'Provisioner ID is required');
 
-    this.influx = new base.stats.Influx(config.influx);
+    if (config.influx.connectionString) {
```
gregarndt: nice to have this spit out to the console when debugging rather than seeing the connection errors when testing locally.  Could change this to emit somethign on process instead and then we could listen for those and test the stats as well...

deploy/packer/base/scripts/packages.sh
```diff
@@ -18,7 +18,7 @@ sudo sh -c "echo deb https://get.docker.io/ubuntu docker main\
 sudo apt-get update -y
 
 ## Install all the packages
-sudo apt-get install -y lxc-docker btrfs-tools lvm2 curl build-essential \
+sudo apt-get install -y lxc-docker-1.6.1 btrfs-tools lvm2 curl build-essential \
```
gregarndt: pin version for now, subsequent PR coming soon will be pinning this, the vagrant image, and CI image to >=1.7.0

bin/worker.js
```diff
@@ -136,8 +136,8 @@ co(function *() {
     config[field] = program[field];
   });
 
-  // If isolated containers is set override capacity...
-  if (config.isolatedContainers) {
+  // If isolated containers is set override capacity (as long as capacity is > 0
+  if (config.isolatedContainers && config.capacity > 0) {
```
gregarndt: if for some reason the host config sets capacity to zero, dont' override it with the isolated containers option

jonasfj: You want this because hosts sets `capacity = 0` to provoke shutdown... where would it do so before this point?
In respawns? Ahh, when the spot shutdown thing is being advertised and we restart... I don't like that we have that restart, but hey... I guess this is safe...

IMO, we should avoid the respawning, the problem with doing it is that this is somewhat obscure...

I would write a longer comment explaining why you do this.

lib/stats/stat.js
jonasfj: I guess this is okay... but really can't we just make a hack in `base.stats` to support doing this? Maybe if connectionString = `null` then we print with debug?

lib/task_listener.js
jonasfj: Definitely log something with `[alert-operator]`. this shouldn't happen... And it's rather bad if it gets deployed like this.

PR #135, opened 2015-07-17
Docker save squashed

EggyLv999:
One more try to get it to go through ci


gregarndt:
So it does appear that perhaps this is a scope issue since the image tar is a private artifact.:

{
  "message": "Authorization Failed",
  "error": {
    "info": "None of the scope-sets was satisfied",
    "scopesets": [
      "queue:get-artifact:private/dockerImage.tar"
    ]
  }
}


worker-ci tests do not have scopes to retrieve public artifacts because as it stands now we only have two options...grant scope to all private artifacts, or just to a specific one that coudl be the same for even non-ci tasks (such as private/dockerImage.tar).  Jonas and I discussed namespacing the private images, something like 'private/worker-ci/dockerImage.tar' or 'private/veryRestrictedAccess/dockerImage.tar'.  This could get a little tricky I think because you would only be able to access private saved images (perhaps also only allowed to create them) if you have scopes to that image to begin with.

Also, I don't think every save needs to be private?  What about a public image that did a clone, ran some tests and otherwise is generally public?

Here are some suggestions:
image saves are public by default unless
   A: explicitly marked as private
   B: private image (task.scopes should include docker-worker:image:<image>)

Also, image saves default to [public|private]/dockerImage.tar unless explicitly overridden in the feature payload:
features: {
  dockerSave: {
    artifactPath: 'public/garndts_image_0720.tar'
  }
}

This way I do not have to grant queue:get-artifact:private/dockerImage.tar to our test credentails which could potentially download the saved images from private tasks, but limit it to queue:get-artifact:private/worker-ci/*

Just some thoughts.

EggyLv999:
The PR now passes CI, because I changed private to public. Unfortunately, there is no way to name images yet, as features does not take non-boolean arguments (attempting to make it do so results in some really, really wonky errors). However, the current form is functional on its own, as long as the user knows that the image is public so they don't hide any private keys or environment variables in there.

test/integration/docker_save_test.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    assert(result.run.state === 'completed', 'task should be successful');
+    assert(result.run.reasonResolved === 'completed',
+                 'task should be successful');
+
+    let taskId = result.taskId;
+    let runId = result.runId;
+
+    let signedUrl = worker.queue.buildSignedUrl(
```
gregarndt: Since this is no longer a private artifact, it doesn't need a signed URL.  We have a helper for retrieving artifacts if you wish to use it:

https://github.com/taskcluster/docker-worker/blob/adbff89b2a7910756197b99c666aba4006e1d122/test/integration/helper/get_artifact.js

gregarndt: Not that it matters much here, but if you want to reduce a call to the queue, and since we no longer need a signed url, you could just build up the artifact url here without an external call.
```js
let url = `https://queue.taskcluster.net/v1/task/${taskId}/runs/${runId}/artifacts/public/dockerImage.tar`
```
            taskId + '/runs/' + runId + '/artifacts/' + path;

gregarndt:
This is looking good to me.  After this lands, the feature should be fully documented on docs.taskcluster.net as we do the other features.

gregarndt:
If you're good with this, we can merge it.

lib/upload_to_s3.js
```diff
@@ -0,0 +1,48 @@
+import https from 'https';
+import url from 'url';
+
+export default async function uploadToS3 (task, sourceStream, artifactName, expires, httpsHeaders) {
+  var queue = task.runtime.queue;
+
+  // Create date when this artifact should expire (see config).
+  var expiration =
```
gregarndt: Could it be possible to just pass in a Date object for expires and let the caller calculate what the expiration should be for the artifact instead of passing in a millisecond value?

You could then assert that expires is an instance of Date before continuing.

lib/features/docker_save.js
```diff
@@ -0,0 +1,48 @@
...
+    let zipStream = zlib.createGzip();
+    imgStream.pipe(zipStream).pipe(fs.createWriteStream(pathname));
+    await waitForEvent(zipStream, 'end');
+    debug('tar written');
+
+    let stat = await fs.stat(pathname);
+    debug(stat.size);
+    let uploadStream = fs.createReadStream(pathname);
+
+    await uploadToS3(task, uploadStream, ARTIFACT_NAME, 60 * 1000, {
```
gregarndt: Wouldn't this make the artifact available for only 1 minute?

gregarndt:
Looking at the graph from this, and the changes made, I think we're good with merging this if you want.

PR #136, opened 2015-07-17
Update docker to 1.7 and pass cpuset as string

gregarndt:


lib/devices/cpu_device_manager.js
```diff
@@ -51,12 +51,15 @@ export default class CpuDeviceManager {
 
 class CpuDevice {
   constructor(id, active=false) {
-    this.id = id
+    this.id = id;
```
jonasfj: Not at all important.. .but I would either do the string conversion here.
Or even better `assert(typeof(id) === 'string')`

jonasfj: I assumed you added the string conversion because it created serious issues when `id` wasn't a string... And this is easily missed in the future.

lib/stats/stat.js
```diff
@@ -15,12 +15,17 @@ export default class Stat {
...
+    // should be a modification to taskcluster-base
     if (config.influx.connectionString) {
       this.influx = new base.stats.Influx(config.influx);
     }
     else {
       this.influx = {
-        addPoint: (...args) => { debug('stats: %j', args); }
+        addPoint: (...args) => { debug('stats: %j', args); },
+        close: () => { return; }
       };
```
jonasfj: The es6 way of doing this:
```js
this.influx = {
  addPoint(...args) { debug('stats: %j', args); },
  close() {}
};
```
Just saying ES6 cool, don't bother fixing it :)

lib/task.js
```diff
@@ -224,8 +224,8 @@ export default class Task {
     };
 
     // Zero is a valid option so only check for existence.
-    if ('cpuset' in this.options) {
-      procConfig.create.Cpuset = this.options.cpuset;
+    if ('cpusetCpus' in this.options) {
+      procConfig.create.HostConfig.CpusetCpus = this.options.cpusetCpus;
```
jonasfj: You renamed this to match docker 1.7.1 I imagine...

jonasfj:
`r+`.

lib/task.js
gregarndt: Yea they are deprecating the other option so figured now is a good time to switch

lib/devices/cpu_device_manager.js
gregarndt: In 1.7.1 they are unmarshalling to go types and before having a number didn't break but now it throws an error that a number cannot be unmarshalled to a string

gregarndt: I actually am thinking of writing a docker config scheme validator

lib/stats/stat.js
gregarndt: Oh I much prefer the cool way, thanks!

lib/devices/cpu_device_manager.js
jonasfj: > I actually am thinking of writing a docker config scheme validator
As a JSON schema? you publish somewhere, that would be awesome...
Because then it could be used in different languages and applications...

PR #139, opened 2015-07-21
Bug 1185848 fail artifact upload

gregarndt:


jonasfj:
Doesn't this resolve the task as `exception` rather than `failed`. I think failed is more appropriate, because this is the same as not exiting zero... Ie. not exiting in a state that it should..

-------
Note, it would be faster to review this if the refactoring to await was done separately...
It's fine for now... Just something to remember next time... Or maybe just an encouragement to go through all the old code and refactor what you want.

gregarndt:
It will be marked as a failure (we don't have an exception reason that suits this purpose anyways currently). https://github.com/taskcluster/docker-worker/blob/master/lib/task.js#L703 (although the lines above it with the logging statement are incorrect since it won't be reported as an exception.  The tests were updated to check to make sure it's marked as failed, not exception as well.


Sorry about the refactoring grouped in this PR, and I'll make sure future PRs will not include refactoring as well.  Figured since I was touching this code and running the tests specific to it, that it was a good time to just make the smaller changes from migrating away from yield/co var/let, etc

PR #141, opened 2015-07-24
Bug 1184272: fix scope matching in docker-worker

djmitche:


lib/task.js
```diff
@@ -77,17 +78,8 @@ Create a list of cached volumes that will be mounted within the docker container
...
-  for (let volumeName in taskVolumeBindings) {
-    neededScopes.push('docker-worker:cache:' + volumeName);
-
-    if (!scopeMatch(taskScopes, neededScopes)) {
-      throw new Error(
-        'Insufficient scopes to attach "' + volumeName + '" as a cached ' +
-        'volume.  Try adding ' + neededScopes + ' to the .scopes array.'
-      );
-    }
+  if (!hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes) {
```
gregarndt: should be if (!hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes)) {

djmitche: fixed

lib/task.js
```diff
@@ -125,10 +117,12 @@ function runAsPrivileged(task, allowPrivilegedTasks) {
 }
 
 function buildDeviceBindings(devices, taskScopes) {
+  if (!hasPrefixedScopes('docker-worker:capability:device:', devices, taskScopes) {
```
gregarndt: Sorry, same problem here...

djmitche:
Yikes, sorry.  I can't tell if the unit tests are hanging or just really slow.. but they didn't bomb out with a syntax error, so I'm figuring this is in better shape now.

lib/task.js
```diff
@@ -77,17 +78,8 @@ Create a list of cached volumes that will be mounted within the docker container
...
-    neededScopes.push('docker-worker:cache:' + volumeName);
-
-    if (!scopeMatch(taskScopes, neededScopes)) {
-      throw new Error(
-        'Insufficient scopes to attach "' + volumeName + '" as a cached ' +
-        'volume.  Try adding ' + neededScopes + ' to the .scopes array.'
-      );
-    }
+  if (!hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes)) {
+    throw new Error('Insufficient scopes to attach cache volumes.');
```
gregarndt: so the nice thing that was happening before was giving an error message in the logs about what exactly to add to resolve this issue.  Here we are losing that.

djmitche: Agreed.  It wasn't quite "nice" since it didn't give the right set of scopes, though.  How about adding something like "the task must have scope `docker-worker:cache:<cache-name>` for each cache in `payload.caches`."?  Given that information, it's trivial for a user to figure out what scopes they need.

gregarndt: Fantastic.  Thank you.

lib/task.js
```diff
@@ -11,6 +11,7 @@ import States from './states';
 
 import features from './features';
 import getHostname from './util/hostname';
+import hasPrefixedScopes from './util/scopes';
```
gregarndt: I looked into the test failures and this needs to be:
import { hasPrefixedScopes } from './util/scopes';

or in util/scopes.js:14 you make it like this:
export default function hasPrefixedScopes(prefix, resources, taskScopes) {

lib/task.js
gregarndt: Also, whatever we include in the log should be updated in these tests:
https://github.com/taskcluster/docker-worker/blob/master/test/integration/container_volume_caching_test.js#L253

djmitche: done

lib/task.js
djmitche: done

gregarndt:
Hrm, seems the CI stuf is having problems, and there is a merge conflict.  Could you rebase against master and we'll see if we can get these tests kicked off.

djmitche:
The conflict is with 85e7b97d85a21d858e6e61e5c586d9b268e967af

djmitche:
Sorry, bad copy/paste -- with whitespace in fec8927d4a2e0eb85535ac2da01249ed162c4c9d

djmitche:
rebased and squashed

gregarndt:
I took a look at chunk 2 failures, and I think the following might clear it up.  The failures in chunk 4 are unfortunately expected right now.

https://gist.github.com/gregarndt/193e3e4b9044345722e4

djmitche:
cherry-picked, thanks

gregarndt:
With that latest cherry pick, the tests are green where I would expect them to be.  If you're ok with these changes now, I can merge.

djmitche:
ship it!

gregarndt:
\o/

PR #140, opened 2015-07-22
Interactive rebase

EggyLv999:
taskcluster-ci senpai please notice me
Also rebased on the current master branch with a bunch of merge conflicts fixed.

PR #144, opened 2015-07-25
Interactive 1.7

EggyLv999:
Adding the docker 1.7 specific changes to combat random failing tests in chunk 4. This PR Merges #136 and #140.


EggyLv999:
@gregarndt No idea why the random test in chunk 4 is failing. Everything else passes. Any ideas/help me investigate please?

gregarndt:
Sure, let me check it out later today or Monday morning to see if I can figure out why.  I'm not sure what might have changed to cause the intermittent but it does seem to happen in this patch and I haven't seen it in the other PRs we've merged.  Perhaps a weird coincidence as well.  Not ruling that one out.

PR #145, opened 2015-07-27
Cache save

EggyLv999:
Cache save feature


EggyLv999:
All tests expected to pass are passing. Needs review.

lib/features/cache_save.js
```diff
@@ -0,0 +1,53 @@
...
+let debug = Debug('docker-worker:features:cache-save');
+
+//must be changed if the delimimter in lib/volume_cache.js is changed
+let KEY_DELIMITER = '::';
+
+export default class DockerSave {
+  async killed (task) {
+    // var packStream = tarStream.pack();
+    await Promise.all(task.volumeCaches.map(async (cacheStr) => {
+      let cache = cacheStr.split(KEY_DELIMITER);
```
jonasfj: Cache keys can be split... but this is not the intention... It's a bad idea... we should add to an extra return value to `buildVolumeBindings` and patch `task.js` to attach the list of `cacheInstance` returned by `volumeCache.get(volumeName);`. And the use the `cacheInstance.path` property instead..

At the moment this creates a situation where we can't change how cache keys are constructed... Ideally, they should just have been uuids.

lib/features/cache_save.js
```diff
@@ -0,0 +1,53 @@
...
+      let loc = path.join(task.runtime.cache.volumeCachePath, cache[0], cache[1] + '/');
+
+      //temporary path for saved file
+      let pathname = path.join('/tmp', slugid.v4() + '.tar.gz');
+      let zipStream = tar.pack(loc, { dereference: true }).pipe(zlib.createGzip());
+      zipStream.pipe(fs.createWriteStream(pathname));
+      await waitForEvent(zipStream, 'end');
```
jonasfj: Could `error` not happen? what if we run out of diskspace or something?

lib/features/cache_save.js
```diff
@@ -0,0 +1,53 @@
...
+      
+      let expiration = new Date(Date.now() + task.runtime.dockerSave.expiration);
```
jonasfj: A better way is to put a string in `dockerSave.expiration` and use `taskcluster.fromNow()` or `taskcluster.fromNowJSON()` see `taskcluster-client` for details...

lib/features/cache_save.js
```diff
@@ -0,0 +1,53 @@
...
+      });
+
+      debug('%s uploaded', cache[0]);
+
+      //cleanup
+      fs.unlink(pathname).catch(() => {
+        task.runtime.log('[alert-operator] could not delete cache save tarball, worker may run out of hdd space');
+      });
+    })).then(() => { debug('all cache(s) uploaded'); }, (e) => {
+      debug('cache could not be uploaded: ' + e + e.stack);
```
jonasfj: Bad idea... If one of the fails, then `Promise.all` is resolved... and as soon as this happens `killed()` returns so cache folder can be used by another task.

You must instead do:
```js
var errors = [];
await Promise.all(volumeCaches.map(async (volume) => {
  try {
    ...
  } catch (err) {
    errors.push(err);
  }
}));

if (errors.length >= 0) {
  .... // probably fail the task... 
}
```

lib/features/cache_save.js
```diff
@@ -0,0 +1,53 @@
...
+          'content-type': 'application/x-tar',
+          'content-length': stat.size,
+          'content-encoding': 'gzip'
+      });
+
+      debug('%s uploaded', cache[0]);
+
+      //cleanup
+      fs.unlink(pathname).catch(() => {
+        task.runtime.log('[alert-operator] could not delete cache save tarball, worker may run out of hdd space');
```
jonasfj: This log the error too...

lib/features/cache_save.js
```diff
@@ -0,0 +1,53 @@
...
+      let zipStream = tar.pack(loc, { dereference: true }).pipe(zlib.createGzip());
+      zipStream.pipe(fs.createWriteStream(pathname));
+      await waitForEvent(zipStream, 'end');
+      
+      let expiration = new Date(Date.now() + task.runtime.dockerSave.expiration);
+      let stat = await fs.stat(pathname);
+
+      await uploadToS3(task,
+        fs.createReadStream(pathname),
+        'public/' + cache[0] + '.tgz',
```
jonasfj: How about: `public/cache/<cacheName>.tar.gz`

jonasfj:
Note, I think we should make this part of the `dockerSave` feature...
Basically, nobody should want caches without the docker container and the other way around...
You don't have the file system state without both of these.

gregarndt:
Definitely like the idea Jonas proposes about combining docker save and cache save.

lib/features/cache_save.js
EggyLv999: Agreed, but

EggyLv999:
As requested, I've merged the two features.

lib/features/docker_save.js
```diff
@@ -9,41 +11,100 @@ import zlib from 'zlib';
...
-    debug('artifact uploaded');
-
-    //cleanup
-    fs.unlink(pathname).catch(() => {
-      task.runtime.log('[alert-operator] could not delete docker save tarball, worker may run out of hdd space');
-    });
-    await image.remove();
+    let errors = [];
+    let volumeCaches = task.volumeCaches || [];
+    await Promise.all(volumeCaches.map(async (cacheStr) => {
```
jonasfj: Small thing... but if you have two min.. maybe refactor into two methods:
```js
 await Promise.all(
  volumeCaches
    .map(cacheStr => this.uploadCache(task, cacheStr).catch(err => error.push(err)))
    .concat(this.uploadContainer(task).catch(err => error.push(err)))
)
```
Then you won't need the `try/catch` inside either `uploadCache` or `uploadContainer` and the code will be a little easier to read.. Especially, if you write a comment documenting the two methods like:
```js
  /** Upload a cache folder as an artifact */
  async uploadCache(task, cacheStr) {
    ....
  },
  /** Commit and upload task container as an artifact */
  async uploadContainer(task) {
    ....
  }
```

lib/features/docker_save.js
```diff
@@ -9,41 +11,100 @@ import zlib from 'zlib';
...
+        await new Promise((accept, reject) => {
+          zipStream.on('end', accept);
+          zipStream.on('error', (error) => reject(error));
+        });
+        let expiration = taskcluster.fromNow(task.runtime.dockerSave.expiration);
+        let stat = await fs.stat(pathname);
+
+        await uploadToS3(task,
+          fs.createReadStream(pathname),
+          'public/cache/' + cache[0] + '.tar.gz',
```
jonasfj: My bad... when I think I told you to use `.tar.gz`, I saw you changes in the test and realize `.tar` is sufficient because when using `content-encoding: gzip` then the HTTP layer (curl, wget, superagent, etc) will do the gzip decompression for you. It's suddenly acts as a transport layer encoding instead of being the encoding of our downloaded result..

@gregarndt, ^? crazy or reasonably sane... I think we do this with logs too... We don't append `.gz` because the HTTP client handles gzip when S3 serves with `content-encoding: gzip`

lib/features/docker_save.js
```diff
@@ -9,41 +11,100 @@ import zlib from 'zlib';
...
+        await uploadToS3(task, uploadStream, 'public/dockerImage.tar.gz', expiration, {
+          'content-type': 'application/x-tar',
+          'content-length': stat.size,
+          'content-encoding': 'gzip'
+        });
+
+        debug('docker image uploaded');
+
+        //cleanup
+        fs.unlink(pathname).catch(() => {
```
jonasfj: `.catch(err => ...`

lib/features/docker_save.js
gregarndt: We create a gzip file for the logs.  we pipe the task stream to zlib.createGzip(), and then that to a write stream to a local file.  The logs then get uploaded (with the .gz extension) and the content-encoding is set.  I think clients when pulling down files with that content-encoding, they will strip the gz and decode it?

lib/features/docker_save.js
```diff
@@ -23,26 +27,81 @@ export default class DockerSave {
...
     debug('tar written');
 
     let stat = await fs.stat(pathname);
     debug(stat.size);
     let uploadStream = fs.createReadStream(pathname);
-    let expiration = new Date(Date.now() + task.runtime.dockerSave.expiration);
+    let expiration = taskcluster.fromNow(task.runtime.dockerSave.expiration);
 
-    await uploadToS3(task, uploadStream, ARTIFACT_NAME, expiration, {
+    await uploadToS3(task, uploadStream, 'public/dockerImage.tar.gz', expiration, {
```
jonasfj: As discussed before please strip `.gz` we don't need it when `content-encoding: gzip` is set...

lib/features/docker_save.js
```diff
@@ -23,26 +27,81 @@ export default class DockerSave {
...
+    await new Promise((accept, reject) => {
+      zipStream.on('end', accept);
+      zipStream.on('error', (error) => reject(error));
+    });
+    let expiration = taskcluster.fromNow(task.runtime.dockerSave.expiration);
+    let stat = await fs.stat(pathname);
+
+    await uploadToS3(task,
+      fs.createReadStream(pathname),
+      'public/cache/' + cache[0] + '.tar.gz',
```
jonasfj: As discussed before please strip `.gz` we don't need it when `content-encoding: gzip` is set...

test/integration/docker_save_test.js
```diff
@@ -45,7 +46,7 @@ suite('use docker-save', () => {
     let taskId = result.taskId;
     let runId = result.runId;
 
-    let url = `https://queue.taskcluster.net/v1/task/${taskId}/runs/${runId}/artifacts/public/dockerImage.tar`;
+    let url = `https://queue.taskcluster.net/v1/task/${taskId}/runs/${runId}/artifacts/public/dockerImage.tar.gz`;
```
jonasfj: strip .gz

lib/features/docker_save.js
```diff
@@ -23,26 +27,81 @@ export default class DockerSave {
...
+    fs.unlink(pathname).catch((err) => {
+      task.runtime.log('[alert-operator] could not delete cache save tarball, worker may run out of hdd space\r\n'
+        + err + err.stack);
     });
     await image.remove();
   }
+
+  //uploading a cache folder as an artifact
+  async uploadCache(task, cacheStr) {
+    let cache = cacheStr.split(KEY_DELIMITER);
```
gregarndt: so this doesn't absolutely need to be changed.. but perhaps an alternative to putting the key delimiter and logic for determining the path in here, a method could be added to the volume cache manager that retrieves the path based on the key.

lib/features/docker_save.js
```diff
@@ -23,26 +27,81 @@ export default class DockerSave {
...
+    let loc = path.join(task.runtime.cache.volumeCachePath, cache[0], cache[1] + '/');
+
+    //temporary path for saved file
+    let pathname = path.join('/tmp', slugid.v4() + '.tar');
```
gregarndt: We might want to look at storing this somewhere under /mnt.  The AWS instances are created with 8gb root volumes, and then 80+GB mounted under /mnt.

Here is a sample from one of our instances:
$ df -H
/dev/xvda1                        8.4G  1.8G  6.1G  23% /
none                              4.1k     0  4.1k   0% /sys/fs/cgroup
udev                              7.9G   13k  7.9G   1% /dev
tmpfs                             1.6G  1.2M  1.6G   1% /run
none                              5.3M     0  5.3M   0% /run/lock
none                              7.9G  1.3M  7.9G   1% /run/shm
none                              105M     0  105M   0% /run/user
/dev/mapper/instance_storage-all   80G   22G   54G  29% /mnt

lib/features/docker_save.js
```diff
@@ -23,26 +27,81 @@ export default class DockerSave {
...
+        + err + err.stack);
+    });
+  }
+
+  async killed(task) {
+    let errors = [];
+    let volumeCaches = task.volumeCaches || [];
+
+    await Promise.all(volumeCaches.map((cacheStr) => this.uploadCache(task, cacheStr).catch(err => error.push(err)))
+    .concat(this.uploadContainer(task).catch((err) => errors.push(err))));
```
gregarndt: Is there no value in separating out saving and uploading errors?  Here it seems that all errors will be treated as upload errors right?

lib/features/docker_save.js
jonasfj: Do we have a configuration jey for where to store stuff like this... It's probably an issue with dockerSave in it's current form too...

lib/features/docker_save.js
```diff
@@ -23,26 +27,81 @@ export default class DockerSave {
...
+      zipStream.on('end', accept);
+      zipStream.on('error', (error) => reject(error));
+    });
+    let expiration = taskcluster.fromNow(task.runtime.dockerSave.expiration);
+    let stat = await fs.stat(pathname);
+
+    await uploadToS3(task,
+      fs.createReadStream(pathname),
+      'public/cache/' + cache[0] + '.tar',
+      new Date(Date.now() + 60 * 60 * 1000), {
```
gregarndt: so the caches are hardcoded to being around for an hour but the docker containers will be around for 7 days (default) ?

If that's intentional we probably want to add a comment why so it's not changed accidentally.

lib/features/docker_save.js
gregarndt: well I stored it here, but perhaps the name could be different:

https://github.com/taskcluster/docker-worker/blob/master/config/defaults.js#L85

test/integration/docker_save_test.js
```diff
@@ -47,10 +48,12 @@ suite('use docker-save', () => {
 
     let url = `https://queue.taskcluster.net/v1/task/${taskId}/runs/${runId}/artifacts/public/dockerImage.tar`;
 
-    //superagent means no unzipping required
+    //superagent means no zlib required
     let res = await request.get(url).end();
     res.pipe(fs.createWriteStream('/tmp/dockerload.tar'));
     await waitForEvent(res, 'end');
+    //make sure it finishes unzipping
+    await base.testing.sleep(1000);
```
gregarndt: Is there some kind of race from superagent emitting the 'end' event and the file being available?

EggyLv999: I think so. I'm not really sure how to fix it otherwise.

lib/features/docker_save.js
EggyLv999: I think in the long run, we might want to consider saving the bindings as an array of objects with 
`
{name: test-cache
slugid: ofdxyug043ntlky32
}
`
as entries.

jonasfj: Yes, and to facilitate that we should make volumeCaches into a list of objects. Ideally use the objects returned by cache manager...

lib/features/docker_save.js
EggyLv999: All errors will be treated as errors, and ideally they will contain information about whether they are save or upload related within the errors themselves. IE a docker errors vs an HTTPS error should be pretty clear which is from which.

EggyLv999:
@gregarndt @jonasfj thoughts? Is this ready to merge?

gregarndt:
it's a small nit, but if it's not too much hassle, I would like to see the path being returns by the volume cache manager rather than just asking it to split a key on some delimiter.  This way getting the location and constructing the path lives within the cache manager and no logic needs to be elsewhere.  Otherwise this is looking very good.  \o/

lib/features/docker_save.js
```diff
@@ -9,12 +11,14 @@ import zlib from 'zlib';
 
 let debug = Debug('docker-worker:features:docker-save');
 
-const ARTIFACT_NAME = 'public/dockerImage.tar'
+//must be changed if the delimiter in lib/volume_cache.js is changed
+let KEY_DELIMITER = '::';
```
gregarndt: I think this is unused now.

lib/features/docker_save.js
```diff
@@ -23,26 +24,81 @@ export default class DockerSave {
...
+    let cache = task.runtime.volumeCache.splitKey(cacheStr);
+    let loc = task.runtime.volumeCache.getCacheLocation(cacheStr);
+
+    //temporary path for saved file
+    let pathname = path.join(task.runtime.dockerVolume, slugid.v4() + '.tar');
+    let zipStream = tar.pack(loc, { dereference: true }).pipe(zlib.createGzip());
+    zipStream.pipe(fs.createWriteStream(pathname));
+    await new Promise((accept, reject) => {
+      zipStream.on('end', accept);
+      zipStream.on('error', (error) => reject(error));
```
jonasfj: just do `zipStream.on('error', error);`...

lib/features/docker_save.js
```diff
@@ -23,26 +24,81 @@ export default class DockerSave {
     let imgStream = await image.get();
     let zipStream = zlib.createGzip();
     imgStream.pipe(zipStream).pipe(fs.createWriteStream(pathname));
-    await waitForEvent(zipStream, 'end');
+    await new Promise((accept, reject) => {
+      zipStream.on('end', accept);
+      zipStream.on('error', (error) => reject(error));
```
jonasfj: nit... `zipStream.on('error', reject);`

EggyLv999:
For some reason, the docker-save test fails a lot due to docker itself. There are two main issues: 1. docker decides that somewhere between save/commit and load, the repo name is lost and 2. sometimes it just randomly freezes and doesn't cat properly, timing out instead. I'll look into 1. more later tonight, but 2. is really confusing me.

jonasfj:
@EggyLv999,
Make sure it's not running out of disk space...

Also check that garbage collector (in docker-worker) doesn't remove the docker image as you save it.
That would certainly explain it... Maybe it removes the image while you're saving it or something like that...

Btw, when you say freezing, do you mean it just hangs... It could be an error we forget to catch resulting a promise that isn't resolved.

jonasfj:
Also check that `_` and `-` are allowed in image names.. That could explain why it fails sometimes... I think they are allowed, but it certainly would explain something.

jonasfj:
I suspect image name `task/Da_lWMzORmulxFWyjRQc-A/0:latest` isn't the best.
Maybe `task-Da_lWMzORmulxFWyjRQc-A-run-0:latest` is better, we have to match `[a-zA-Z0-9-_.]` I think.

slugids have `_` and `-` occasionally, so testing that it works with both of them makes sense..

EggyLv999:
Things fixed: Cannot find named image

Errors still occurring: 
 - Corrupted tar, docker somehow loads it in just fine but it doesn't have a name, causing the test to fail because it can't find the image
 - Docker stream doesn't output anything in test: this one is technically ok because the docker image is actually downloaded just fine and can be used, but for some reason the attached stream just never outputs
 - Freeze: this one is super rare and I haven't really had a chance to investigate it. Probably an awaited promise that never returns, I have no idea what causes this.

EggyLv999:
Turns out that besides the naming issue, the other issues were all due to docker making images out of almost-complete files (these tars were all 1KB below average) because my own /mnt folder is apparently writable, but not really? Anyway, the other failures were all due to running on local and using /mnt; moving to /tmp fixed it for local and shouldn't affect test machines.

lib/volume_cache.js
```diff
@@ -222,4 +222,17 @@ export default class VolumeCache {
       this.cache[cacheName][instanceId][key] = value[key];
     }
   }
+
+  /**
+  Get the name and location of a cache on disk.
+
+  @param {String} Cache key in the format of <cache name>::<instance id>
+  */
+  splitKey(cacheKey) {
```
gregarndt: nit: the naming of this does not really represent what it's doing.  Perhaps getCacheDetails or something.

gregarndt:
I think this is looking good and I'm ok with it being merged

PR #147, opened 2015-07-27
Key delimiter refactor

EggyLv999:
moving the key delimiter option to config so it can be accessed in multiple places.

### DIFF
--- a/config/test.js
+++ b/config/test.js
@@ -32,7 +32,7 @@ module.exports = {
   },
 
   cache: {
-    volumeCachePath: path.join(__dirname, '..', 'test', 'tmp')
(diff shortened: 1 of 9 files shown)


config/test.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
   },
 
   cache: {
-    volumeCachePath: path.join(__dirname, '..', 'test', 'tmp')
+    volumeCachePath: path.join('/tmp', 'test-cache')
```
gregarndt: is the reason for polluting things outside of the docker-worker directory so that we have a determinate path for tests?

EggyLv999: This was me trying to fix failing tests that really shouldn't have been. I'll take a closer look at this patch later.

PR #153, opened 2015-08-01
Increasing the poll interval for workers nearing end of billing cycle

EggyLv999:
I'm not sure how much of a change this will have, but I do know this:

On a workertype with 1 worker (of which there are very few), we are at worst adding 70 seconds of wait time. On a workertype with many workers, this will hopefully reduce the number of workers sitting around with low used capacity by allowing workers to pull in larger batches of tasks at once. Of course, this is speculation, because I don't know what changing the polling from like an average of 20 polls/second for a big workertype like b2gtest or flame-kk to 3 polls/second will do. It also slows polling after a while, allowing workers nearing the end of a billing cycle to have a higher chance of dying instead of picking up new tasks. Maybe it won't do much. Hopefully though, this lets workers shut down when they should be shutting down, maybe reducing our AWS bill by a smidgeon.

### DIFF
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -220,7 +220,7 @@ async function main () {
 
   var runtime = new Runtime(config);
 
-  runtime.host = host;
+  runtime.hostManager = host;
 
   // Instantiate PrivateKey object for decrypting secure data
   // (currently encrypted environment variables)
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -219,7 +219,7 @@ async function main () {
   config.gc.addManager(config.volumeCache);
 
   var runtime = new Runtime(config);
-  
+
   runtime.host = host;
 
   // Instantiate PrivateKey object for decrypting secure data
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -240,7 +240,7 @@ async function main () {
   }
 
   // Build the listener and connect to the queue.
-  var taskListener = new TaskListener(runtime);
+  var taskListener = new TaskListener(host, runtime);
   runtime.gc.taskListener = taskListener;
   shutdownManager.observe(taskListener);
 
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -219,6 +219,8 @@ async function main () {
   config.gc.addManager(config.volumeCache);
 
   var runtime = new Runtime(config);
+  
+  runtime.host = host;
 
   // Instantiate PrivateKey object for decrypting secure data
   // (currently encrypted environment variables)
@@ -240,7 +242,7 @@ async function main () {
   }
 
   // Build the listener and connect to the queue.
-  var taskListener = new TaskListener(host, runtime);
+  var taskListener = new TaskListener(runtime);
   runtime.gc.taskListener = taskListener;
   shutdownManager.observe(taskListener);
 
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -127,7 +127,11 @@ module.exports = {
 
   taskQueue: {
     // Task queue will be polled on a frequent interval for new pending tasks
-    pollInterval: 30 * 1000,
+    pollInterval: 5 * 1000,
+    // Task queue will be polled slower near the end of a billing cycle by this many times
+    pollIntervalMultiplier: 12,
+    // Task polling will slow down in the last 1/x of a billing cycle
+    slowdownDivisor: 3,
     // If signed url for queue expires within now()+expiration, refresh queues
     expiration: 5 * 60 * 1000,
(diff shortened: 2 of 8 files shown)


bin/worker.js
```diff
@@ -240,7 +240,7 @@ async function main () {
   }
 
   // Build the listener and connect to the queue.
-  var taskListener = new TaskListener(runtime);
+  var taskListener = new TaskListener(host, runtime);
```
gregarndt: I would suggest adding the host manager to the runtime in case later on this needs to be used elsewhere.

config/defaults.js
```diff
@@ -128,6 +128,10 @@ module.exports = {
   taskQueue: {
     // Task queue will be polled on a frequent interval for new pending tasks
     pollInterval: 5 * 1000,
+    // Task queue will be polled slower near the end of a billing cycle by this many times
+    pollIntervalMultiplier: 12,
```
gregarndt: Perhaps add to the comment as to why these are not just magic numbers to use.

lib/task_listener.js
```diff
@@ -181,6 +182,18 @@ export default class TaskListener extends EventEmitter {
               stack: e.stack
           });
         }
+        let stats = {
+          uptime: this.host.billingCycleUptime(),
+          interval: this.host.billingCycleInterval()
+        };
+        let remainder = stats.interval - (stats.uptime % stats.interval);
+        if(remainder * this.runtime.taskQueue.slowdownDivisor > stats.interval) {
```
gregarndt: we should add something to the task polling test to verify this logic and to make sure it doesn't accidentally get changed later on.

EggyLv999: @gregarndt Any ideas on how to do this? The uptime and interval can't really be adjusted for testing purposes, so I'm not really sure how to test this without waiting for 40 minutes for the slowdown period to begin on its own.

gregarndt: Sure, wait 40 minutes!  No problem in that..

or you could do something like:
https://github.com/taskcluster/docker-worker/blob/master/test/integration/shutdown_test.js#L31
https://github.com/taskcluster/docker-worker/blob/master/test/integration/shutdown_test.js#L13

lib/states.js
```diff
@@ -60,7 +60,7 @@ export default class States {
       {state: method}
     ).then(results => {
       if (errors.length > 0) {
-        throw new Error(errors.map(e => e.toString()).join(' | '))
+        throw new Error(errors.map(e => (e.toString() + e.stack)).join(' | '));
```
gregarndt: I'm not sure if the stack trace should be included here.  These errors get caught in task.js and end up in the log.  We should log these errors and stack traces to debug and remove things like this: https://github.com/taskcluster/docker-worker/blob/master/lib/task.js#L585

test/integration/task_polling_test.js
```diff
@@ -97,4 +98,43 @@ suite('Task Polling', () => {
...
+    settings.billingCycleUptime(30);
+    settings.billingCycleInterval(40);
+
+    worker = new TestWorker(DockerWorker);
+    await worker.launch();
+    //make sure the polling timer is accurate
+    await base.testing.sleep(6000);
+
+    settings.billingCycleUptime(0);
+    await new Promise((accept, reject) => {
```
gregarndt: So one of the things we've done in tests is using the wait_for_event module and you can do something like

let pollingMessage = await waitForEvent(worker, 'polling');
assert.equal(pollingMessage.message, 'polling interval adjusted', <some error message if this fails>)

Either this happens or the test will time out.

test/integration/task_polling_test.js
```diff
@@ -97,4 +98,43 @@ suite('Task Polling', () => {
...
+      worker.on('polling', data => {
+        console.log(data.message);
```
gregarndt: Also, regardless if we use waitForEvent or this promise, we don't want to be logging to console.log in tests and clouding up the test output.

lib/states.js
EggyLv999: But aren't the stack traces of these errors lost forever when we use e.toString()? Or am I misunderstanding something here?

lib/task_listener.js
```diff
@@ -181,6 +182,32 @@ export default class TaskListener extends EventEmitter {
               stack: e.stack
           });
         }
+        let stats = {
+          uptime: this.host.billingCycleUptime() || 0,
```
gregarndt: If the only reason to do this is because the tests don't specify it, I would rather find a way to include in the tests rather than defaulting here.

PR #154, opened 2015-08-04
Hopefully fixes issues with querystring in interactive

EggyLv999:


EggyLv999:
If there's no complaints, this should be merged asap so I can check tools integration on interactive.

PR #155, opened 2015-08-05
Bug 1063110: switch to base.validator.

mihneadb:
r? @gregarndt @jonasfj

### DIFF
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -173,6 +173,9 @@ async function main () {
   config.scheduler =
     new taskcluster.Scheduler({ credentials: config.taskcluster });
 
+  config.validator = await base.validator();
+  config.validator.register(require('../schemas/payload'));
+
   // Only catch these metrics when running on aws host.  Running within
   // test environment causes numerous issues.
   if (host === 'aws') {
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -172,7 +172,6 @@ async function main () {
 
   config.scheduler =
     new taskcluster.Scheduler({ credentials: config.taskcluster });
-  config.schema = require('../lib/schema')();
 
   // Only catch these metrics when running on aws host.  Running within
   // test environment causes numerous issues.
--- a/lib/task.js
+++ b/lib/task.js
@@ -598,7 +598,7 @@ export default class Task {
       return await this.abortRun(this.taskState);
     }
     // Validate the schema!
-    let payloadErrors = await validator().then(function (validator) {
+    let payloadErrors = await validator().then((validator) => {
       return validator.check(this.task.payload, payloadSchema);
     });
 
--- a/lib/task.js
+++ b/lib/task.js
@@ -5,6 +5,7 @@ import Debug from 'debug';
 import DockerProc from 'dockerode-process';
 import util from 'util';
 import uuid from 'uuid';
+import validator from 'base';
 import { PassThrough } from 'stream';
 import States from './states';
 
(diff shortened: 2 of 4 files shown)


mihneadb:
Used `push --force` to clean up the history a bit. Should work now.

lib/task.js
```diff
@@ -599,10 +598,11 @@ export default class Task {
       return await this.abortRun(this.taskState);
     }
     // Validate the schema!
-    let payloadErrors =
-      this.runtime.schema.validate(this.task.payload, PAYLOAD_SCHEMA);
+    let payloadErrors = await validator().then(function (validator) {
+      return validator.check(this.task.payload, payloadSchema);
```
EggyLv999: Try ```js
() => {return validator.check(this.task.payload, payloadSchema);
}```
in place of the anonymous function. As it stands, (I believe) the contextual `this` may be set to the wrong thing.

gregarndt: Ah, it might be.  using arrow functions now makes this a lot less painful as it always captures the context of the surrounding scope.

mihneadb: Forgot about the gotcha, thanks! :+1:

mihneadb:
@gregarndt seems ok now, pls take another look. Thanks!

gregarndt:
This is looking good to me and tests pass.  @jonasfj things look good to you? I haven't used the validator before, but from what I've read it seems sane.

mihneadb:
@gregarndt good to hear, thanks. After this gets merged I'll fix the other bug (default cmd args).

bin/worker.js
```diff
@@ -172,7 +172,6 @@ async function main () {
 
   config.scheduler =
     new taskcluster.Scheduler({ credentials: config.taskcluster });
-  config.schema = require('../lib/schema')();
```
jonasfj: This should be something like:
```js
config.validator = await base.validator();
config.validator.register(require('../schemas/payload')); // or whereever its...
```
This way we only create the validator object once... And in a future where we have more than one schema we can load them globally...

lib/task.js
```diff
@@ -13,14 +14,12 @@ import getHostname from './util/hostname';
...
 import { pullDockerImage, IMAGE_ERROR } from './pull_image_to_stream';
 import { scopeMatch } from 'taskcluster-base/utils';
+import payloadSchema from '../schemas/payload';
 import waitForEvent from './wait_for_event';
 import _ from 'lodash';
 
 let debug = new Debug('runTask');
 
-const PAYLOAD_SCHEMA =
-  'http://schemas.taskcluster.net/docker-worker/v1/payload.json#';
```
jonasfj: Probably need to keep this. see comment below...

lib/task.js
```diff
@@ -599,10 +598,11 @@ export default class Task {
       return await this.abortRun(this.taskState);
     }
     // Validate the schema!
-    let payloadErrors =
-      this.runtime.schema.validate(this.task.payload, PAYLOAD_SCHEMA);
+    let payloadErrors = await validator().then((validator) => {
```
jonasfj: See change above about global validator instance... then this becomes:
let payloadErrors = this.runtime.validator.check(this.task.payload, PAYLOAD_SCHEMA);

jonasfj:
Using a global instance of `validator` is much smarter... especially if we have multiple schemas... But also necessary in a future where we perhaps have a more expensive compilation step converting JSON schemas into JSON code... the `is-my-json-valid` module does this... and I want to do that in `base.validator` in some future :)

mihneadb:
@jonasfj I've addressed your comments.

jonasfj:
Looks good to me... @gregarndt, are you merging?

PR #157, opened 2015-08-06
Bug 1190660: fixing expiration dates on feature artifacts

EggyLv999:
fixes part of this bug by cutting expiration dates down to task.expires at most.

### DIFF
--- a/lib/features/azure_live_log.js
+++ b/lib/features/azure_live_log.js
@@ -51,8 +51,6 @@ export default class LiveLog {
   async created(task) {
     var queue = task.runtime.queue;
 
-    console.log(task.task.expires ? new Date(task.task.expires) : Infinity)
-
     // Create date when this artifact should expire (see config).
     var expiration = new Date(
       Math.min(Date.now() + task.runtime.logging.liveLogExpires,
--- a/lib/features/azure_live_log.js
+++ b/lib/features/azure_live_log.js
@@ -51,9 +51,12 @@ export default class LiveLog {
   async created(task) {
     var queue = task.runtime.queue;
 
+    console.log(task.task.expires ? new Date(task.task.expires) : Infinity)
+
     // Create date when this artifact should expire (see config).
-    var expiration = new Date(Math.min(Date.now() + task.runtime.logging.liveLogExpires,
-      task.task.expires ? new Date(task.task.expires) : Infinity));
+    var expiration = new Date(
+      Math.min(Date.now() + task.runtime.logging.liveLogExpires,
+      new Date(task.task.expires)));
 
     var options = {
       storageType: 'azure',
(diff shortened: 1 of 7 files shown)


lib/features/azure_live_log.js
```diff
@@ -52,8 +52,8 @@ export default class LiveLog {
     var queue = task.runtime.queue;
 
     // Create date when this artifact should expire (see config).
-    var expiration =
-      new Date(Date.now() + task.runtime.logging.liveLogExpires);
+    var expiration = new Date(Math.min(Date.now() + task.runtime.logging.liveLogExpires,
+      task.task.expires ? new Date(task.task.expires) : Infinity));
```
jonasfj: When would `task.expires` be missing? We get the task from `queue.task` response schema says that it's always present...
The queue inserts a default if it's missing :)

jonasfj: Note, two lines might be preferable for readability:
```js
var expiration = new Date(Math.min(
  Date.now() + task.runtime.logging.liveLogExpires,
  new Date(task.task.expires)
));
```

jonasfj:
@EggyLv999, you can make an attachment on the bug just provide the URL for the github PR as patch... and then file for a r?...
That way we get this linked from the bug, and there is a slighter higher chance we remember to resolve bug :)

gregarndt:
Looks like the tests pass, but we should add a test for ensuring that the live log is created when a task expires in say, a day? and that the expiration on that artifact is correct.

test/integration/logging/local_live_log_test.js
```diff
@@ -78,6 +76,23 @@ suite('live logging', () => {
       token.length,
       22, 'Token should be 22 characters long'
     );
+  });
 
+  test('live log expiration date', async () => {
+    let expiration = taskcluster.fromNowJSON('1 second');
```
gregarndt: Hrm, is this sufficiently enough time for the task to be submitted and claimed consistently? anything less than one year would work since the default is now() + 1 year for the default expiration.

test/integration/logging/local_live_log_test.js
```diff
@@ -78,6 +76,23 @@ suite('live logging', () => {
...
+          '/bin/bash',
+          '-c',
+          'echo "first command!"'
+        ],
+        maxRunTime: 3 * 60
+      }
+    });
+    assert(result.artifacts['public/logs/live.log'].expires === expiration,
+      'expiration date of live log improperly set');
   });
```
gregarndt: we probably also want to assert the task was successful

test/integration/logging/local_live_log_test.js
EggyLv999: Fun fact: live logs for tests were expiring in 3.6 seconds instead of 3600 seconds=1 hour. This was a workaround that I forgot to figure out after I got the tests working.

gregarndt:
Great, tests passed, changes look good to me.

PR #158, opened 2015-08-06
fixed typo s/successfull/successful

EggyLv999:
simple typo fix, hopefully it doesn't break anything


gregarndt:
Many thanks for this.  Might only be spelling corrections, but even those are pretty embarrassing.  Docker-worker is definitely better off with these fixes, so thank you :)

PR #160, opened 2015-08-08
1098460 default cmd

mihneadb:
Fixes bug 1098460.

Depends on #155. Those parts of the diff will disappear when #155 is merged.

TODO after merge:
- [ ] publish new schema

r? @gregarndt @jonasfj

### DIFF
--- a/schemas/payload.js
+++ b/schemas/payload.js
@@ -46,7 +46,7 @@ module.exports = {
   "title": "Docker worker payload",
   "description": "`.payload` field of the queue.",
   "type": "object",
-  "required": ["image", "command", "maxRunTime"],
+  "required": ["image", "maxRunTime"],
   "properties": {
     "image": {
       "title": "Docker image.",
@@ -118,10 +118,10 @@ module.exports = {
     "command": {
       "title": "Docker command to run (see docker api).",
       "type": "array",
-      "minItems": 1,
       "items": {
         "type": "string"
       },
+      "default": [],
       "description": "Example: `['/bin/bash', '-c', 'ls']`."
     },
     "encryptedEnv": {
--- a/lib/task.js
+++ b/lib/task.js
@@ -141,7 +141,7 @@ function buildDeviceBindings(devices, taskScopes) {
   return deviceBindings;
 }
 
-export default class Task {
+export class Task {
   /**
   @param {Object} runtime global runtime.
   @param {Object} task id for this instance.
--- a/lib/task.js
+++ b/lib/task.js
@@ -19,7 +19,7 @@ import _ from 'lodash';
 
 let debug = new Debug('runTask');
 
-const PAYLOAD_SCHEMA =
+export const PAYLOAD_SCHEMA =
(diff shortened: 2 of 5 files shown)


gregarndt:
This, and the other PR are looking good to me, I think we could probably be safe merging this.  Just asking @jonasfj for another set of eyes on the schema validation pieces.  I appreciate the help on this!

mihneadb:
Sweet, no problem!

schemas/payload.js
```diff
@@ -118,10 +118,10 @@ module.exports = {
     "command": {
       "title": "Docker command to run (see docker api).",
       "type": "array",
-      "minItems": 1,
       "items": {
         "type": "string"
       },
+      "default": [],
```
jonasfj: Yeah, this looks right to me :)

jonasfj:
problems. in PR #155 should be addressed first... Mainly use a global validator instance.

mihneadb:
@jonasfj do you still see the purpose of the separate module that I was planning on unit testing? Since we only use one validator and that is from an external library, we shouldn't test it in this repo.
We should probably still test that the schema does what we expect it to do, so a minimal set of unit tests sound useful to me. What do you think?

jonasfj:
> We should probably still test that the schema does what we expect it to do, so a minimal set of unit tests sound useful to me. What do you think?

Testing that the schema does what we expect is nice...
An added bonus is that if we change schema in the future, the errors from these tests will let us know if we broke backward compatibility.

So yeah, a minimal task set is perhaps sane. Since initialize `base.validator` is still so simple we can perhaps just duplicate that into the test case. I'm not entirely sure how to do it right, but I do something similar in other places were I test schemas.
`base.testing.schemas` can load schemas from a folder and test against a set of cases. But it won't work here because the schema is a javascript file that generates it, when loaded... So we have to use `require`.

mihneadb:
@gregarndt @jonasfj I've updated the PR as discussed. Please review.

jonasfj:
looks good to me...
@gregarndt, I hope you're into clicking the merge button :)
Assuming you have no objections.

mihneadb:
Also don't forget to publish the new schema.

gregarndt:
I'll try to get some time this weekend to check out the errors in the task graph, but it seems that tests are having a hard time now.

mihneadb:
@gregarndt you're saying there is some unrelated aspect that made the tests fail?

gregarndt:
It looks like the change to export the payload schema in task.js broke the import in task_listener.js.  I think this should fix it up:

https://github.com/gregarndt/docker-worker/commit/e58795adad699cf941297dc475b2d1916cf0362a

mihneadb:
Oh wow :(. Should I cherry-pick that commit and try again?

gregarndt:
Whatever works for you :).  However you add it, also make sure to rebase against master in case things have changed since then.

mihneadb:
@gregarndt done! Thanks!

gregarndt:
This looks much better now, thanks!

PR #161, opened 2015-08-10
Reclaim errors should be caught and task aborted

gregarndt:


lib/features/local_live_log.js
```diff
@@ -164,6 +164,14 @@ export default class TaskclusterLogs {
 
   async killed(task) {
     debug('switching live log redirect to backing log...')
+    try {
+      await this.container.stop();
```
jonasfj: Shouldn't we just force kill the livelog container?

jonasfj: And what did we do before?
Did we just leave it for the GC... so that it could finish serving existing connections?

jonasfj:
`r+`

lib/features/local_live_log.js
gregarndt: Before the container would just keep running until it was explicitly marked to be removed.

PR #164, opened 2015-09-02
Move ssl cert and keys to base AMI creation

gregarndt:
Goal here is to move the certificates and papertrail setup to the base image so regenerating the app ami does not require any secrets.  

Some options for some of the secret storing involved:
1. using something like blackbox that hasn't been proven for our uses and no other team uses it
2. storing encoded string in provisioner UI (this would be preferred as it's the source of all secrets for a worker, but makes the worker definition extremely long when 3 keys need to be in there)
3. papertrail logging endpoint needs to be setup before anything else so we don't lose syslog messages in papertrail

Moving these pieces that do not change often at all (private key and papertrail endpoint hasn't changed since first introduced) to the base image frees up the work that needs to be done to create an app ami and anyone now with access to our account can generate an app ami easily.


deploy/packer/base.json
```diff
@@ -1,12 +1,41 @@
...
+      "source": "{{user `privateKeyLocation`}}",
+      "destination": "/tmp/docker-worker-priv.pem"
+    },
+    { "type": "file",
+      "source": "{{user `sslCertificateLocation`}}",
+      "destination": "/tmp/star_taskcluster-worker_net.crt"
+    },
+    { "type": "file",
+      "source": "{{user `sslKeyLocation`}}",
+      "destination": "/tmp/star_taskcluster-worker_net.key"
```
jonasfj: why not call it `worker-ssl.key`... if someday TC is deployed outside tc.net context it would be weird...

Anyways, not you don't have to fix it :)

jonasfj:
looks good to me... `r+`

I suspect point (3) is the reason we don't encrypt these on the AMI and store encryption key in provisioner... 

FYI, I don't see big issues from leaking these, you would still have to compromise DNS to do anything fun with SSL key...

gregarndt:
Yea I should have put that thought in my comment, I did consider going the route of the provisioner have a key in the secrets to decrypt the stuff, but then that would just mean instead of 3 keys in the worker definition we would have one, which is still less than ideal.

deploy/packer/base.json
gregarndt: I'll leave it for now, because if we did deicde to deploy outside of TC someday ( I hope!) there is a bunch of other work needed to be done to the worker to make that a reality.  Best to just tackle it all at once.

jonasfj:
>  I did consider going the route of the provisioner have a key in the secrets to decrypt the stuff, but then that would just mean instead of 3 keys in the worker definition we would have one, which is still less than ideal.

Symmetric keys can be 32 bytes... they are much smaller... and easy to fit into provisioner secrets...

jonasfj:
anyways, for papertrail it would be not fun, because we want to to run from start... for SSL keys it wouldn't be that hard... but no need yet..

jonasfj:
these things are hard to abuse anyways...

PR #165, opened 2015-09-05
Improve error handling within features

gregarndt:


lib/task.js
```diff
@@ -742,29 +742,16 @@ export class Task {
...
-          stack: e.stack
-        });
-
-        this.stream.write(this.fmtErrorLog(
-          `Unknown taskcluster error encountered.  Ask administrator to lookup ` +
-          `incidentId in log-file. Incident ID: ${lookupId}`
-        ));
       }
+
+      this.stream.write(this.fmtErrorLog(e.message));
```
jonasfj: I don't like this... we should really spec how plugins/features work... and what expected... Errors should not be a way of communicating error messages about cases we handle expectedly...

Errors should be for things going bad... This is not bad, just bad input. Two dramatically different things...

jonasfj: Note, this might be acceptable sort term...

gregarndt: I absolutely agree. This is not really changing the behavior of how we communicate errors but just improving the error messages in the interim. what I would like to do is have a spec for methods features implement and the type of states they can fail with (such as exception or failure). I also want to try to communicate more informative errors rather than dumping whatever exception was thrown from wherever. That almost never helps the end user.

gregarndt:
So I would like to release this in the short term until we can get around to refactoring how features work and implement 'internal-error' and 'resource-unavailable' which will need us to change how errors are reported back from features (failures vs exceptions)

jonasfj:
Fair enough...

PR #168, opened 2015-09-24
Image artifact

gregarndt:


lib/docker/docker_image.js
```diff
@@ -0,0 +1,191 @@
+import Debug from 'debug';
```
gregarndt: Most of this is a combination of lib/docker_image.js and lib/pull_image_stream_to.js

lib/docker/image_manager.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return this._lastImageEnsured = Promise.resolve(this._lastImageEnsured)
+      .catch(() => {}).then(async () => {
+        this.log('ensure image', {
+         image: imageDetails
+        });
+
+        let imageHandler = this.getImageHandler(imageDetails, stream, scopes);
+        let exists = await imageHandler.imageExists();
```
jonasfj: why make a new object every time this is confusing...

Instead why not have two lists, `this._dockerImages` and `this._indexedImages`?
Then for each image only have one instance of the image object.

lib/docker/image_manager.js
```diff
@@ -0,0 +1,95 @@
...
+   */
+  getImageHandler(image, stream, scopes) {
+    let handler;
+    if (image.type === 'docker-image') {
+      handler = new DockerImage(this.runtime, image, stream, scopes);
+    } else if (image.type === 'indexed-image') {
+      handler = new IndexedImage(this.runtime, image, stream, scopes);
+    } else {
+      throw new Error(`Unrecognized image type. Image Type: ${image.type}`);
+    }
```
jonasfj: So good way to structure this code is to have a base-class (even if it just documents methods), dictionary of strategies in `image_manager.js`.. Oh, an instead of `imageHandler.download` I would just make a factory.. Example:
```js
class ImageHandler extends events.EventEmitter {
  constructor() {
    this._refCount = 0;
  }
  /** gets the imageId that it's inserted with in docker */
  get _dockerImageId { throw new Error("Must be implement in subclass"); }
  /* determine if scopes grants access to the image */
  _isAuthorized(scopes) { throw new Error("Must be implement in subclass"); }
  garbageCollect() {
    if (this._refCount === 0) {
      this._remove();
      this.emit('deleted', this);
    }
  }
  /** delete resources held by this image */
  _remove() {throw new Error("Must be implement in subclass");}
  /**Use this image in `func(dockerImageId)` which may be an async function,
   * this way you never forget to decrement refCount */
  async use(func) {
    this._refCount += 1;
    try {
      await func(this._dockerImageId)
    } finally {
      this._refCount -= 1;
    }
  }
};
```

```js

const IMAGE_LOADERS = {
  'docker-image': require('./docker-image'),  // these files then export an async function that load the image 
  'indexed-image': require('./tar-image').indexedLoader, // and after loading creates a subclass of ImageHandler
  'artifact-image': require('./tar-image').artifactLoader, // that they then return...
};

class ImageManager {
  ...
  ensureImage(imageInfo, stream, scopes) {
    var id = this.hash(imageInfo);
    if (this._cache[id]) {
      return this._cache[id];
    }
    // So cache[id] is a promise, hence, two subsequent calls will never download the image twice
    // or construct the ImageHandler object twice.
    this._cache[id] = this._lastImageEnsured = Promise.resolve(this._lastImageEnsured).then(() => {
       return IMAGE_LOADERS[imageInfo.type](imageInfo, stream, scopes);
    });
    this._cache[id].then(handler => {
      handler.once('deleted', h => {
        delete this._cache[id]; // remove from cache when it's is deleted
      });
    });
    return this._cache[id];
  }
  hash(imageInfo) { /* compute hash from imageInfo */ }
};
```

```js
class TarImageHandler extends ImageHandler {
  ....
};

exports.indexedLoader = async (...) => {
  // look up taskId of indexed artifact
  exports.artifactLoader(taskId, .....); // So reuse the other logic
};

exports.indexedLoader = async (...) => {
   // download, do the tar stream things... 
  return new TarImageHandler(...);
};
```
This is just illustrative.. The notable thing is that `ImageHandler` subclasses can't have errors. They don't do anything async... They just represent a resources and guard access and garbage collection.
`ImageManager` ensures two images aren't loaded at the same time, and that there is only ever one `ImageHandler` for every image.

I think this is a better OOP design... I generally find that whenever you are about to make an async contructor, and find that an async initializer like `download()` is necessary, it's often better to just do a factory method, so all methods on the object wrapping the resource are safe.

Also I don't believe you handle scopes for cached images correctly at the moment. Ie. once image is loaded, you don't really check scopes for the next task trying to use the image. My guess is that `docker-worker` will need to know what scopes it used to download the image artifact...  I think that's acceptable as it's always the same pattern.

PR #172, opened 2015-09-29
Bug 1171809: Implement purge cache events.

walac:
Let a build A creates a cached object directory Ca. Now let a build B
that reuses the cache Ca. If the build A fail and corrupts Ca, build B
will fail as well.

Enters the purge cache service. You can send a purge cache message to
clobber a build.

This commit implements the listening side of purge cache. docker-worker
listens for the purge cache event and remove the named cache. If the
cache is currently mounted, it schedules it for removal on gc container
removed event.

### DIFF
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -62,6 +62,31 @@ function o() {
   program.option.apply(program, arguments);
 }
 
+async function listenForPurgeCacheEvents(runtime, volumeCache) {
+  // see task_listener.js, listenForCancelEvents() comment
+  if (!runtime.pulse || !runtime.pulse.username || !runtime.pulse.password) {
+    runtime.log('[alert-operator] pulse credentials missing');
+    return;
+  }
+
+  let purgeCacheEvents = new taskcluster.PurgeCacheEvents();
+  let purgeCacheListener = new taskcluster.PulseListener({
+    credentials: runtime.pulse
+  });
+
+  await purgeCacheListener.bind(purgeCacheEvents.purgeCache({
+    workerType: runtime.workerType,
+    provisionerId: runtime.provisionerId
+  }));
+
+  purgeCacheListener.on('message', function (message) {
+    volumeCache.purge(message.payload.cacheName);
+  });
+
+  await purgeCacheListener.resume();
+  return purgeCacheListener;
+}
+
 // Usage.
 program.usage(
 '[options] <profile> \n\n' +
@@ -226,6 +251,8 @@ async function main () {
   // (currently encrypted environment variables)
   runtime.privateKey = new PrivateKey(runtime.dockerWorkerPrivateKey);
 
+  runtime.purgeCacheListener = listenForPurgeCacheEvents(runtime, config.volumeCache);
+
   // Billing cycle logic is host specific so we cannot handle shutdowns without
   // both the host and the configuration to shutdown.
   if (host && config.shutdown) {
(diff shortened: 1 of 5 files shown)


test/integration/container_volume_caching_test.js
```diff
@@ -1,3 +1,7 @@
+async function sleep(duration) {
```
gregarndt: I believe that this lives under lib/util/sleep.js

test/integration/container_volume_caching_test.js
```diff
@@ -354,4 +360,59 @@ suite('volume cache tests', function () {
...
+        clientId: process.env.TASKCLUSTER_CLIENT_ID,
+        accessToken: process.env.TASKCLUSTER_ACCESS_TOKEN
+      }
+    });
+
+    purgeCache.purgeCache(worker.provisionerId, worker.workerType, {
+      cacheName: cacheName
+    });
+
+    yield sleep(2000);
```
gregarndt: so I think there are ways around introducing sleeps here.  Basically I think what you're trying to wait for is that the removal actually happened.  There are times where it might take longer than 2 seconds for a pulse message to be sent, received by the worker, and the directory removed.

gregarndt: you could do 
```js
yield waitForEvent(worker, 'cache volume removed')
```

bin/worker.js
```diff
@@ -212,10 +237,12 @@ async function main () {
 
   config.volumeCache = new VolumeCache(config);
 
-  config.gc.on('gc:container:removed', function (container) {
+  config.gc.on('gc:container:removed', async function (container) {
     container.caches.forEach(async (cacheKey) => {
       await config.volumeCache.release(cacheKey);
     });
+
+    await config.volumeCache.doPurge();
```
gregarndt: Hrm, does this need to be done here?  I think that all of the purging stuff could just happen when the GC cycle starts.  the GC could call the volume manager and it can clear up whatever it needs to, including caches that should be purged.  As long as caches marked as "purge: true" don't get reused for a task, I don't see the harm in just deferring it to GC

lib/volume_cache.js
```diff
@@ -235,4 +238,30 @@ export default class VolumeCache {
...
+   Mark cache for purge and try to remove it.
+
+   @param cacheName {String} Name of the cache.
+   */
+  async purge(cacheName) {
+    _.forOwn(this.cache[cacheName], (instance) => {
+     instance.purge = true;
+    });
+
+    await this.doPurge();
```
gregarndt: I'm debating internally about this.  As my other comment suggested, perhaps just leaving this up for the next GC cycle isn't a bad thing and you don't have to worry here about removing anything.  As long as it gets marked for purging, then it can be deferred.

bin/worker.js
```diff
@@ -62,6 +62,31 @@ function o() {
   program.option.apply(program, arguments);
 }
 
+async function listenForPurgeCacheEvents(runtime, volumeCache) {
+  // see task_listener.js
```
gregarndt: What exactly are we seeingin task_listener.js?

gregarndt:
Looks like there are some errors in CI

lib/volume_cache.js
```diff
@@ -120,6 +122,8 @@ export default class VolumeCache {
   @param {Boolean} Disksapce threshold reached
   */
   async clear(exceedsDiskspaceThreshold) {
+    this.doPurge();
```
gregarndt: you should probably wait for this to be done before moving on.

PR #177, opened 2015-10-28
Bug 1186619: Retry artifact upload.

walac:
When an artifact upload fails, we try to upload a limited number of
times before failing it.

### DIFF
--- a/lib/upload_to_s3.js
+++ b/lib/upload_to_s3.js
@@ -1,45 +1,101 @@
 import https from 'https';
 import url from 'url';
+import fs from 'fs';
+import temporary from 'temporary';
+import promiseRetry from 'promise-retry';
+import _ from 'lodash';
 
-export default async function uploadToS3 (task, sourceStream, artifactName, expiration, httpsHeaders) {
-  var queue = task.runtime.queue;
+export default async function uploadToS3 (
+  task,
+  sourceStream,
+  artifactName,
+  expiration,
+  httpsHeaders,
+  putUrl,
+  httpOptions)
+{
+  let queue = task.runtime.queue;
 
-  var artifact = await queue.createArtifact(
-    task.status.taskId,
-    task.runId,
-    artifactName,
-    {
-      // Why s3? It's currently cheaper to store data in s3 this could easily
-      // be used with azure simply by changing s3 -> azure.
-      storageType: 's3',
-      expires: new Date(Math.min(expiration, new Date(task.task.expires))),
-      contentType: httpsHeaders['content-type']
+  let tmp = new temporary.File();
+
+  try {
+    await new Promise((accept, reject) => {
+      let stream = fs.createWriteStream(tmp.path);
+      stream.on('error', reject);
+      stream.on('finish', accept);
+      sourceStream.pipe(stream);
+    });
+
+    if (!putUrl) {
+      var artifact = await queue.createArtifact(
+        task.status.taskId,
+        task.runId,
+        artifactName,
+        {
+          // Why s3? It's currently cheaper to store data in s3 this could easily
+          // be used with azure simply by changing s3 -> azure.
+          storageType: 's3',
+          expires: new Date(Math.min(expiration, new Date(task.task.expires))),
+          contentType: httpsHeaders['content-type']
+        }
+      );
+
+      putUrl = artifact.putUrl;
     }
-  );
 
-  let parsedUrl = url.parse(artifact.putUrl);
-  let options = {
-    hostname: parsedUrl.hostname,
-    path: parsedUrl.path,
-    method: 'PUT',
-    headers: httpsHeaders
-  };
+    var parsedUrl = url.parse(putUrl);
+    var options = _.defaults({
+      hostname: parsedUrl.hostname,
+      path: parsedUrl.path,
+      method: 'PUT',
+      headers: httpsHeaders,
+      port: parsedUrl.port
+    }, httpOptions);
 
-  let req = https.request(options);
-  sourceStream.pipe(req);
+    // promiseRetry defaults to 10 attempts before failing
+    await promiseRetry((retry, number) => {
+      if (number > 1) { // if it's not the first attempt
+        task.runtime.log('retrying artifact upload', {
+          taskId: task.status.taskId,
+          runId: task.runId
+        });
+        task.runtime.log(`Attempt number ${number}`);
+      }
 
-  let response = await new Promise((accept, reject) => {
-    req.on('response', accept);
-    req.on('error', reject);
-    req.setTimeout(5 * 60 * 1000, reject);
-  });
+      return new Promise((accept, reject) => {
+        let req = https.request(options);
 
-  // Flush the data from the reponse so it's not held in memory
-  response.resume();
+        req.on('response', (response) => {
+          // Flush the data from the reponse so it's not held in memory
+          response.resume();
 
-  if (response.statusCode !== 200) {
-    throw new Error(
-      `Could not upload artifact. ${response.error} Status Code: ${response.statusCode}`
-    );
+          if (response.statusCode !== 200) {
+            reject(new Error(
+              `Could not upload artifact. Status Code: ${response.statusCode}`
+            ));
+          } else {
+            accept();
+          }
+        });
+
+        req.on('error', err => {
+          task.runtime.log(`Error uploading ${artifactName}`, {
+            taskId: task.status.taskId,
+            runId: task.runId
+          });
+          reject(err);
+        });
+
+        req.setTimeout(5 * 60 * 1000, reject);
+        task.runtime.log(`Uploading ${artifactName}`, {
+          taskId: task.status.taskId,
+          runId: task.runId
+        });
+        fs.createReadStream(tmp.path).pipe(req);
+      }).catch(retry);
+    // randomize the timeouts
+    }, {randomize: true});
+  } finally {
+    tmp.unlink();
   }
 }
--- a/test/integration/directory_artifact_test.js
+++ b/test/integration/directory_artifact_test.js
@@ -3,8 +3,22 @@ import testworker from '../post_task';
 import getArtifact from './helper/get_artifact';
 import cmd from './helper/cmd';
 import expires from './helper/expires';
+import TestWorker from '../testworker';
+import DockerWorker from '../dockerworker';
+import iptables from 'iptables';
+import _ from 'lodash';
 
-suite('Directory artifact', () => {
+suite('Directory artifact', function() {
+  teardown(() => {
+    iptables.deleteRule({
+      chain: 'OUTPUT',
+      target: 'REJECT',
+      protocol: 'tcp',
+      dport: 443,
+      sudo: true
+    });
+  });
+
   test('attempt to upload file as directory', async () => {
     let result = await testworker({
       payload: {
@@ -78,5 +92,89 @@ suite('Directory artifact', () => {
     assert.ok(Buffer.byteLength(testHtml) === 1000000,
       'Size of uploaded contents of test.html does not match original.'
     );
+  });
+
+  test('retry directory upload', async () => {
+    // Avoid iptables on local environment
+    if (!process.env.WORKER_CI) {
+      return;
+    }
+
+    this.timeout(480000);
+
+    let ARTIFACT_COUNT = 3;
+
+    let worker = new TestWorker(DockerWorker);
+    await worker.launch();
+
+    var count = 0;
+
+    let rejectConnection = function() {
+      // Delay iptables rejection to catch
+      // in the middle of the upload
+      setTimeout(function() {
+        iptables.reject({
+          chain: 'OUTPUT',
+          protocol: 'tcp',
+          dport: 443,
+          sudo: true
+        });
+      }, 100);
+    }
+
+    let cmdArgs = ['mkdir "/xfoo"'];
+    for (let i = 1; i <= ARTIFACT_COUNT; ++i) {
+      cmdArgs.push(`dd if=/dev/zero of=/xfoo/test${i}.html bs=1 count=${1000000*i}`);
+      worker.once(`Uploading public/dir/test${i}.html`, rejectConnection);
+    }
+
+    worker.on('retrying artifact upload', function() {
+      iptables.deleteRule({
+        chain: 'OUTPUT',
+        target: 'REJECT',
+        protocol: 'tcp',
+        dport: 443,
+        sudo: true
+      });
+
+      ++count;
+    });
+
+    let result = await worker.postToQueue({
+      payload: {
+        image: 'taskcluster/test-ubuntu',
+        command: cmd.apply(this, cmdArgs),
+        features: {
+          localLiveLog: false
+        },
+        artifacts: {
+          'public/dir': {
+            type: 'directory',
+            path: '/xfoo/',
+            expires: expires()
+          },
+        },
+        maxRunTime: 5 * 60
+      }
+    });
+
+    await worker.terminate();
+
+    assert.equal(result.run.state, 'completed', 'task should be successful');
+    assert.equal(result.run.reasonResolved, 'completed', 'task should be successful');
+
+    assert.deepEqual(
+      Object.keys(result.artifacts).sort(),
+      [for (i of _.range(1, ARTIFACT_COUNT+1)) `public/dir/test${i}.html`].sort()
+    );
+
+    for (let i = 1; i <= ARTIFACT_COUNT; ++i) {
+      let testHtml = await getArtifact(result, 'public/dir/test' + i + '.html');
+      assert.ok(Buffer.byteLength(testHtml) === 1000000*i,
+        'Size of uploaded contents of test.html does not match original.'
+      );
+    }
+
+    assert.equal(count, ARTIFACT_COUNT, `Count value is ${count}`);
   });
 });
--- a/test/integration/file_artifact_test.js
+++ b/test/integration/file_artifact_test.js
@@ -3,8 +3,21 @@ import getArtifact from './helper/get_artifact';
 import cmd from './helper/cmd';
 import expires from './helper/expires';
 import testworker from '../post_task';
+import TestWorker from '../testworker';
+import DockerWorker from '../dockerworker';
+import iptables from 'iptables';
 
 suite('artifact extration tests', () => {
+  teardown(() => {
+    iptables.deleteRule({
+      chain: 'OUTPUT',
+      target: 'REJECT',
+      protocol: 'tcp',
+      dport: 443,
+      sudo: true
+    });
+  });
+
   test('extract artifact', async () => {
     let result = await testworker({
       payload: {
@@ -227,5 +240,73 @@ suite('artifact extration tests', () => {
 
     // Missing artifact should have an error...
     assert.equal(result.artifacts['public/my-missing.txt'].storageType, 'error');
+  });
+
+  test('upload retry', async () => {
+    // Avoid iptables on local environment
+    if (!process.env.WORKER_CI) {
+      return;
+    }
+
+    let worker = new TestWorker(DockerWorker);
+    await worker.launch();
+
+    worker.once('Uploading public/xfoo', function() {
+      iptables.reject({
+        chain: 'OUTPUT',
+        protocol: 'tcp',
+        dport: 443,
+        sudo: true
+      });
+    });
+
+    let retry = false;
+
+    worker.on('retrying artifact upload', function() {
+      iptables.deleteRule({
+        chain: 'OUTPUT',
+        target: 'REJECT',
+        protocol: 'tcp',
+        dport: 443,
+        sudo: true
+      });
+
+      retry = true;
+    });
+
+    let result = await worker.postToQueue({
+      payload: {
+        image: 'taskcluster/test-ubuntu',
+        command: cmd(
+          'mkdir /artifacts/',
+          'echo "xfoo" > /artifacts/xfoo.txt',
+          'ls /artifacts'
+        ),
+        features: {
+          localLiveLog: false
+        },
+        artifacts: {
+          'public/xfoo': {
+            type: 'file',
+            expires: expires(),
+            path: '/artifacts/xfoo.txt'
+          },
+        },
+        maxRunTime: 5 * 60
+      }
+    });
+
+    // Get task specific results
+    assert.equal(result.run.state, 'completed', 'task should be successful');
+    assert.equal(result.run.reasonResolved, 'completed', 'task should be successful');
+
+    assert.deepEqual(
+      Object.keys(result.artifacts).sort(), ['public/xfoo'].sort()
+    );
+
+    let xfoo = await getArtifact(result, 'public/xfoo');
+
+    assert.equal(xfoo.trim(), 'xfoo');
+    assert.ok(retry);
   });
 });
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "mz": "^2.0.0",
     "openpgp": "1.2.0",
     "promise": "^7.0.3",
+    "promise-retry": "^0.2.9",
     "proxy": "^0.2.3",
     "readdirrsync": "0.0.3",
     "request": "^2.65.0",
@@ -79,6 +80,7 @@
     "mocha": "^2.1.0",
     "split2": "^0.1.2",
     "taskcluster-task-factory": "^0.6.1",
-    "got": "^5.1.0"
+    "got": "^5.1.0",
+    "iptables": "^0.0.4"
   }
 }
--- a/taskgraph.json
+++ b/taskgraph.json
@@ -31,6 +31,9 @@
               "loopbackAudio": true,
               "loopbackVideo": true
             }
+          },
+          "env": {
+            "WORKER_CI": "1"
           }
         },
         "extra": {
@@ -65,6 +68,9 @@
               "loopbackAudio": true,
               "loopbackVideo": true
             }
+          },
+          "env": {
+            "WORKER_CI": "1"
           }
         },
         "extra": {
@@ -99,6 +105,9 @@
               "loopbackAudio": true,
               "loopbackVideo": true
             }
+          },
+          "env": {
+            "WORKER_CI": "1"
           }
         },
         "extra": {
@@ -133,6 +142,9 @@
               "loopbackAudio": true,
               "loopbackVideo": true
             }
+          },
+          "env": {
+            "WORKER_CI": "1"
           }
         },
         "extra": {
@@ -167,6 +179,9 @@
               "loopbackAudio": true,
               "loopbackVideo": true
             }
+          },
+          "env": {
+            "WORKER_CI": "1"
           }
         },
         "extra": {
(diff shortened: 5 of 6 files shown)


test/integration/file_artifact_test.js
```diff
@@ -228,4 +231,67 @@ suite('artifact extration tests', () => {
...
+        chain: 'OUTPUT',
+        protocol: 'tcp',
+        dport: 443,
+        sudo: true
+      });
+    });
+
+    let retry = false;
+
+    worker.on('retrying artifact upload', function() {
```
gregarndt: This is so awesome, I love docker.

gregarndt: Hrm, will this alter the iptables for the host that it's running on? I wonder if we could do this just for the worker container that starts up.

lib/features/artifacts.js
```diff
@@ -167,7 +167,22 @@ export default class Artifacts {
 
     // Upload all the artifacts in parallel.
     await Promise.all(_.map(artifacts, (value, key) => {
-      return this.uploadArtifact(taskHandler, key, value).catch((err) => {
+      return new Promise((accept, reject) => {
+        let numRetries = taskHandler.runtime.retry.numRetries;
+        let retry = () => {
+          this.uploadArtifact(taskHandler, key, value)
```
gregarndt: So one of the main issues here is handling directory artifact uploads.  This works if it's just a file, but we can also specify a directory.  We wouldn't want to retry each file in a directory if it was already successful.

test/integration/file_artifact_test.js
```diff
@@ -228,4 +231,67 @@ suite('artifact extration tests', () => {
     // Missing artifact should have an error...
     assert.equal(result.artifacts['public/my-missing.txt'].storageType, 'error');
   });
+
+  test('upload retry', async () => {
+    let worker = new TestWorker(DockerWorker);
+    await worker.launch();
+
+    worker.on('task run', function() {
+      iptables.reject({
```
gregarndt: We should clean this up outside of the test in case the test itself fails for some reason we will make sure the iptable rules are cleaned up.

config/defaults.js
```diff
@@ -217,5 +217,10 @@ module.exports = {
     expirationAfterSession: 15 * 60,
     //base path of the socket artifact
     artifactName: 'private/docker-worker/'
+  },
+
+  retry: {
```
gregarndt: I think this should be named to describe what the retry is for. (I hate naming so I don't have a good name here :) )

lib/features/artifacts.js
```diff
@@ -128,7 +128,22 @@ export default class Artifacts {
...
-        await uploadArtifact(taskHandler, stream, entryName, expiry, headers);
+        await new Promise((accept, reject) => {
+          let numRetries = taskHandler.runtime.retry.numRetries;
+          let retry = () => {
+            uploadArtifact(taskHandler, stream, entryName, expiry, headers)
+              .then(accept)
+              .catch((err) => {
+                if (numRetries--) {
+                  taskHandler.runtime.log('retrying artifact upload');
+                  setTimeout(retry, taskHandler.runtime.retry.interval);
```
gregarndt: We should probably be doing the retry with some kind of exponential backoff.  There is an example where we do this for pulling down an image.

jonasfj: with small random jitter added too...

lib/features/artifacts.js
```diff
@@ -128,7 +128,22 @@ export default class Artifacts {
...
       try {
-        await uploadArtifact(taskHandler, stream, entryName, expiry, headers);
+        await new Promise((accept, reject) => {
+          let numRetries = taskHandler.runtime.retry.numRetries;
+          let retry = () => {
+            uploadArtifact(taskHandler, stream, entryName, expiry, headers)
+              .then(accept)
+              .catch((err) => {
+                if (numRetries--) {
+                  taskHandler.runtime.log('retrying artifact upload');
```
gregarndt: We should include what the error was for good measure.  runtime.log takes an event and a object payload for things like error messages.

lib/features/artifacts.js
```diff
@@ -128,7 +128,22 @@ export default class Artifacts {
       };
 
       try {
-        await uploadArtifact(taskHandler, stream, entryName, expiry, headers);
+        await new Promise((accept, reject) => {
+          let numRetries = taskHandler.runtime.retry.numRetries;
+          let retry = () => {
+            uploadArtifact(taskHandler, stream, entryName, expiry, headers)
```
gregarndt: So I'm not 100% why this is working.  My understanding is that once it reads from the stream once, that data is drained from the stream so subsequent calls to upload the artifact and read from the same stream *should* fail because the data isn't there.  It might be because this file is below the watermark threshold for the stream (I think 16k).  Perhaps we could have a test that's uploading a directory containing multiple files of multiple sizes and verifying the contents of those files after upload.

test/integration/file_artifact_test.js
jonasfj: this test doesn't work... you're not reading anything from the stream... So you don't test retries...
To make this work you need to read half the stream and then reset...

This will show you that actually can't retry uploading a stream, as reading from stream has side-effects... And our streams can't seek... So we probably need to:
 A) store in a temporary file
 B) reload from docker container (but only reload the file that failed to upload)

(B) could get complicated...

lib/features/artifacts.js
gregarndt: Ok, so I spoke with Jonas and he left some comments below in the tests.  It's true that reading from the stream should cause the data to be drained from it, so subsequent reads should not work as expected because it's not the data you think it is.  

The reason why the test is passing is explained below (blocking 443 causes createArtifact to fail before it even reads from the stream)

jonasfj: This does the `createArtifact` call, it's unlikely that there is a need to retry that (signed url doesn't expire in 5 min)... but I guess okay to do...

It's probably easiest to do all of this inside `uploadArtifact` and use a temporary file.

test/integration/file_artifact_test.js
gregarndt: So I have debated about storing temp files form the container and just initiating another copy operation.  The downside to storing the temp files is now we have a diskspace overhead of storing those until they're fully uploaded, and also making sure we properly remove them.  However it does allow us to do just one copy to not only calculate the hashes of the files (for headers used later on once implemented), but then retrying upload doesn't have to initiate another copy.

lib/features/artifacts.js
selenamarie: /me notes the https://www.npmjs.com/package/promise-retry and https://www.npmjs.com/package/retry packages, both of which have configurable backoff and jitter

gregarndt: nice.  If we change to using a package for retrying here, we should also create an issue to use the same package elsewhere in docker-worker where retrying is done.  Right now the method for retrying and backoff+jitter was inspired (read: copied) from other taskcluster components that do the same.  We should probably standardize how we're doing this across the board.

selenamarie: :100: :dancers:

lib/features/artifacts.js
jonasfj: Also note `createArtifact´ is retried by tc-client...

test/integration/file_artifact_test.js
jonasfj: you can also decide to just read to a buffer if the file is small than say 200 MB... most files probably are... We are already uploading sequentially anyways, as that is how docker gives us the files...

walac:
I upgraded the patch to do retry in `artifactUpload`, also using [promise-retry](https://www.npmjs.com/package/promise-retry) as suggested by @selenamarie :)

lib/upload_to_s3.js
```diff
@@ -1,45 +1,81 @@
...
-    req.setTimeout(5 * 60 * 1000, reject);
+  let queue = task.runtime.queue;
+
+  var artifact;
+  var parsedUrl;
+  var options;
+
+  let tmp = new temporary.File();
+  await new Promise((accept, reject) => {
+    let stream = fs.createWriteStream(tmp.path);
```
gregarndt: I think that this method should only be responsible for uploading a file, not also storing it somewhere.  We could wrap docker.copy() to save to a temp file, and then here we'll open up a read stream from it.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,81 @@
...
+      if (number > 1) {
+        task.runtime.log('retrying artifact upload');
+      }
+
+      return new Promise(async (accept, reject) => {
+        new Promise((accept, reject) => {
+          if (artifact) {
+            accept();
+          } else {
+            queue.createArtifact(
```
gregarndt: I'm really not sure if we need to retry creating the artifact....last time I look we rarely have to retry the call even a second time, and I think taskcluster-client defaults to 5 retries.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,81 @@
...
+          }
+        }).then(() => {
+          let req = https.request(options);
+
+          req.on('response', (response) => {
+            // Flush the data from the reponse so it's not held in memory
+            response.resume();
+
+            if (response.statusCode !== 200) {
+              reject(Error(
```
gregarndt: when calling reject, I don't think an error object needs to be passed in

lib/upload_to_s3.js
```diff
@@ -1,45 +1,81 @@
...
+              task.runId,
+              artifactName,
+              {
+                // Why s3? It's currently cheaper to store data in s3 this could easily
+                // be used with azure simply by changing s3 -> azure.
+                storageType: 's3',
+                expires: new Date(Math.min(expiration, new Date(task.task.expires))),
+                contentType: httpsHeaders['content-type']
+              }
+            ).then((result) => {
```
gregarndt: If we let taskcluster-client retry the createArtifact call, it simplifies the promise chaining that you had to do here.

lib/upload_to_s3.js
jonasfj: correct :)
The only reason to do this, if the upload attempt was really slow... ie.> 30min, at which point the signature on the put url have expired...

IMO, artifact upload should never take 30min, if it does we have serious issues...

lib/upload_to_s3.js
```diff
@@ -1,45 +1,81 @@
...
-  // Flush the data from the reponse so it's not held in memory
-  response.resume();
+  try {
+    await promiseRetry((retry, number) => {
+      if (number > 1) {
+        task.runtime.log('retrying artifact upload');
+      }
+
+      return new Promise(async (accept, reject) => {
+        new Promise((accept, reject) => {
```
jonasfj: not what goes on here... but I doubt want want to be using `new Promise` except for the actual upload perhaps...

you'll want to do `promiseRetry(async (retry, number) => {`

lib/upload_to_s3.js
jonasfj: IMO, not a bad idea... but yeah, not required... and if you wanted to do it... it would be `new Error(...)`

lib/upload_to_s3.js
walac: Well, I see the temp file as an implementation detail of the retry mechanism, so it makes sense be wrapped inside `uploadArtifact`

lib/upload_to_s3.js
walac: All in all, it makes no harm and testing gets a little bit easier if `createArtifact` is part of the retry process.

lib/upload_to_s3.js
walac: Thanks, forgot the `new` keyword.

lib/upload_to_s3.js
walac: I didn't understand your comment. Anyway, I updated the code to use `await` for `createArtifact`.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,80 @@
...
 
-  // Flush the data from the reponse so it's not held in memory
-  response.resume();
+  try {
+    await promiseRetry((retry, number) => {
+      if (number > 1) {
+        task.runtime.log('retrying artifact upload');
+      }
+
+      return new Promise(async (accept, reject) => {
```
jonasfj: `return new Promise(async (accept, reject) => {`

Poor use of `async` and `new Promise`... You risk leaking exceptions here and hanging around forever :)

Don't use `new Promise` unless strictly necessary (ie. dealing with events).

jonasfj: there is an intro to promises and async here: http://docs.taskcluster.net/learn/

There might be better resources out there... Fair warning; I wrote the tutorial on docs.tc.net

test/integration/file_artifact_test.js
jonasfj: oh, for testing that retries work you create a proxy and upload through the proxy... Then have the proxy break the connection.

You can supply a custom `baseUrl` for the queue object and mock the queue API for testing this. But that might be a bit hard... If you removed the `queue.createArtifact` call from the retry cycle you could give a url (that goes through a local proxy) to the upload function in a unit test.

This kind of thing is hard to test in an integration test. Or you add a config option specifying a proxy to use.. so you can break the proxy... Easiest is to just do the unit test and observe that it retries when you give it an invalid URL :)

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
-    req.on('error', reject);
-    req.setTimeout(5 * 60 * 1000, reject);
+export default async function uploadToS3 (
+  task,
+  sourceStream,
+  artifactName,
+  expiration,
+  httpsHeaders,
+  putUrl,
+  additionalOptions)
```
jonasfj: nit, probably rename to `requestOptions` or `httpOptions` as this is specific to the `https.request` method right?

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
+    }
+
+    var parsedUrl = url.parse(putUrl);
+    var options = extend({
+      hostname: parsedUrl.hostname,
+      path: parsedUrl.path,
+      method: 'PUT',
+      headers: httpsHeaders,
+      port: parsedUrl.port
+    }, additionalOptions);
```
jonasfj: lodash.defaults is pretty good too...
`options = _.defaults({....}, additionalOptions)`

Just reducing a dependency, that's all...

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
+{
+  let queue = task.runtime.queue;
+
+  let tmp = new temporary.File();
+  await new Promise((accept, reject) => {
+    let stream = fs.createWriteStream(tmp.path);
+    stream.on('error', reject);
+    stream.on('finish', accept);
+    sourceStream.pipe(stream);
   });
```
jonasfj: An error could happen even if the file is created right?
Ie if the error comes from the sourceStream...

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
+  sourceStream,
+  artifactName,
+  expiration,
+  httpsHeaders,
+  putUrl,
+  additionalOptions)
+{
+  let queue = task.runtime.queue;
+
+  let tmp = new temporary.File();
```
jonasfj: Seems okay for this... But I see little value from this component as you are just generating a tmp file, always in `/tmp/`, using a `uuid.v4()` or `slugid.v4()` would be safer than `temporary.File()` which uses `Math.random()`:
https://github.com/vesln/temporary/blob/master/lib/generator.js#L32

Nit, though I doubt we'll have collision in the same millisecond.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
-      `Could not upload artifact. ${response.error} Status Code: ${response.statusCode}`
-    );
+        req.on('error', reject);
+        req.setTimeout(5 * 60 * 1000, reject);
+        task.runtime.log(`Uploading ${artifactName}`);
+        fs.createReadStream(tmp.path).pipe(req);
+      }).catch(retry);
+    }, {randomize: true});
+  } finally {
+    tmp.unlink();
```
jonasfj: you should handle errors here and at least log them... I'm not sure it's worth crashing over :)

jonasfj:
Generally, this looks sane to me :)

jonasfj:
I'll never be a fan of testing with iptables (I fear leaving it in a dirty state, granted we can just restart vagrant)... I suspect unit tests would be good enough for testing the retry part of the upload, it certainly be easier that way.

gregarndt:
Definitely can agree with the issues with using iptable hacks for testing this.  While it can get the job done, it's like trying to beat a nail in with a sledgehammer.  Also, it's under the assumption that most will run these tests under a vagrant environment so the risks of something getting screwed up is minimal, but as I've discovered lately, at least in one case, someone was not able to set that environment up on linux and ran tests locally.

lib/upload_to_s3.js
gregarndt: so just as a suggestion, I would look at the work that Dustin is doing in the coalescer PR where he's also touching this method.  Just to prevent you two from clashing on the chnages.

lib/upload_to_s3.js
gregarndt: Well it uses the current time plus math.random().  I would guess the likelihood of a collision being very low (same time down to the ms along with the same psuedo-random number).

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
+  additionalOptions)
+{
+  let queue = task.runtime.queue;
+
+  let tmp = new temporary.File();
+  await new Promise((accept, reject) => {
+    let stream = fs.createWriteStream(tmp.path);
+    stream.on('error', reject);
+    stream.on('finish', accept);
+    sourceStream.pipe(stream);
```
gregarndt: How are we handling if there is an error from the sourcestream, which in most cases is a copy stream from docker?  I don't know if this has happened much, I suspect not.  Perhaps we could just reject with a useful error message that it couldn't copy the file?

test/mocha.opts
```diff
@@ -1,6 +1,6 @@
 --ui tdd
 --require test/helper
---timeout 240s
+--timeout 480s
```
gregarndt: Why are we increasing this here?

lib/upload_to_s3.js
gregarndt: there's also Object.assign() that might work here if you want to use no libraries at all.

```js
Object.assign(
    {},
    {
        hostname: parsedUrl.hostname,
        path: parsedUrl.path,
        method: 'PUT',
        headers: httpsHeaders,
        port: parsedUrl.port
    },
    additionalOptions
);
```

lib/upload_to_s3.js
gregarndt: I agree the likelihood of a collision is very low, but what I worry about is maybe a very rare possibility that temp files are created for two different tasks using the same temp file path, and one of those was meant to be a private artifact and now gets uploaded as a public artifact for another task.  

Granted this is a very, very unlikely case.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
+    var parsedUrl = url.parse(putUrl);
+    var options = extend({
+      hostname: parsedUrl.hostname,
+      path: parsedUrl.path,
+      method: 'PUT',
+      headers: httpsHeaders,
+      port: parsedUrl.port
+    }, additionalOptions);
+
+    await promiseRetry((retry, number) => {
```
gregarndt: nit: could we name this something other than "number"...like "attempts"

gregarndt: I guess this is similar to the documentation for promise-retry so it doesn't need to be changed.  Just seemed like a non-descriptive name to me.

gregarndt: Could we add a comment as to the behavior of promise-retry here, such as why the decision to turn on randomize and that it defaults to 10 retries so it doesn't seem as magical with what it's doing.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
+      }
+
+      return new Promise((accept, reject) => {
+        let req = https.request(options);
+
+        req.on('response', (response) => {
+          // Flush the data from the reponse so it's not held in memory
+          response.resume();
+
+          if (response.statusCode !== 200) {
```
gregarndt: So I wonder if in the cases whre there is an authorization error, do we want to retry upload?  Mostly the issues we're seeing are with errors connectoni resets, timeouts, etc.

test/integration/directory_artifact_test.js
```diff
@@ -79,4 +83,81 @@ suite('Directory artifact', () => {
...
+    let worker = new TestWorker(DockerWorker);
+    await worker.launch();
+
+    var count = 0;
+
+    let rejectConnection = function() {
+      // Delay iptables rejection to catch
+      // in the middle of the upload
+      setTimeout(function() {
+        iptables.reject({
```
gregarndt: so now that I've found out that at least one person will run docker tests locally outside of a vagrant environment, this makes me very scared that this doesn't get cleaned up after a test and all secure connections are then rejected after running the test.  Perhaps there is a way we can find out if this is running in our vagrant environment or automation and run it? I thought that this would be an easy use of iptables magic, but it seems the downsides are starting to pile up.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,84 @@
...
-    req.on('response', accept);
-    req.on('error', reject);
-    req.setTimeout(5 * 60 * 1000, reject);
+export default async function uploadToS3 (
+  task,
+  sourceStream,
+  artifactName,
+  expiration,
+  httpsHeaders,
+  putUrl,
```
gregarndt: I see that we can pass in the putURL now...is this being done anywhere?

lib/upload_to_s3.js
walac: Hm, right, moved the `awit new Promise` stuff inside the `try` statement.

lib/upload_to_s3.js
walac: Things are actually more complicated. Temporary file names suffer from the [TOCTTOU](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use). The only reliable way to generate temporary files is by using the OS corresponding system call, like [tmpfile](http://linux.die.net/man/3/tmpfile) or [GetTempFileName](https://msdn.microsoft.com/en-us/library/aa364991(v=vs.85).aspx). I look at several nodejs packages but no one implemented a reliable temp file function. In summary, anyway we are theoretically screwed.

lib/upload_to_s3.js
walac: I am generally more in favor of letting the caller handling errors and logging, at its has more context.

lib/upload_to_s3.js
walac: The only issue here is let a temporary file hanging in the OS. I fixed it by moving this inside the `try` statement. As I said above, IMO it's better let the caller deals with error handling.

test/mocha.opts
walac: because [this test](https://github.com/taskcluster/docker-worker/pull/177/files#diff-0f281f5ca74802b8893be95ed5839938R87) takes some time because of the big files.

lib/upload_to_s3.js
walac: Ok, done.

lib/upload_to_s3.js
walac: How often does this happen? I mean, if it is a corner case, I wonder if it's worth adding another code path just for the task to fail earlier.

test/integration/directory_artifact_test.js
walac: That's the only way we can get artifacts retry integration tests so far, which is very important, IMO. Although it has its drawbacks, testing I think the using iptables is still worth.

lib/upload_to_s3.js
walac: Ooops, forget to `git add` the new test file. *facepalm*

walac:
Added a new commit with the pointed issues fixed. What I didn't fix, I left a comment for further discussion. This new commits is only to make review easier, I will squash it before merging the code. 
Now time for new Star Wars movie.

lib/upload_to_s3.js
gregarndt: I'm not sure how often, so I suppose we'll just see what happens.

test/integration/directory_artifact_test.js
gregarndt: Hrm, could we make this test disabled by default unless running in automation or something? I just do not want someone running this locally screwing up their ip tables on their host.

test/mocha.opts
gregarndt: Ah, maybe we could find out a way to just set the timeout for this one test.  Maybe there is a suite.timeout() method in mocha rather than enabling it globally.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,87 @@
...
+      path: parsedUrl.path,
+      method: 'PUT',
+      headers: httpsHeaders,
+      port: parsedUrl.port
+    }, httpOptions);
+
+    // promiseRetry defaults to 10 attempts before failing
+    await promiseRetry((retry, number) => {
+      if (number > 1) { // if it's not the first attempt
+        task.runtime.log('retrying artifact upload');
```
gregarndt: Could we add some information here about the task ID, run ID, and artifact name that's being retried?

lib/upload_to_s3.js
```diff
@@ -1,45 +1,87 @@
...
+              `Could not upload artifact. Status Code: ${response.statusCode}`
+            ));
+          } else {
+            accept();
+          }
+        });
+
+        req.on('error', reject);
+        req.setTimeout(5 * 60 * 1000, reject);
+        task.runtime.log(`Uploading ${artifactName}`);
```
gregarndt: Please add the task ID, run ID if possible as well.

lib/upload_to_s3.js
```diff
@@ -1,45 +1,87 @@
...
+          if (response.statusCode !== 200) {
+            reject(new Error(
+              `Could not upload artifact. Status Code: ${response.statusCode}`
+            ));
+          } else {
+            accept();
+          }
+        });
+
+        req.on('error', reject);
```
gregarndt: I wonder if we should log here that there was an error and what the error was.

lib/upload_to_s3.js
gregarndt: Oh, and the retry attempt number might be helpful too.

test/integration/directory_artifact_test.js
```diff
@@ -79,4 +83,86 @@ suite('Directory artifact', () => {
...
+      let worker = new TestWorker(DockerWorker);
+      await worker.launch();
+
+      var count = 0;
+
+      let rejectConnection = function() {
+        // Delay iptables rejection to catch
+        // in the middle of the upload
+        setTimeout(function() {
+          iptables.reject({
```
gregarndt: I thought I left a comment about this previously related to iptables and teardown of a test... but I think somehow this is failing and then leaving around the ip table rule that breaks after tests after that with "Error: connect ECONNREFUSED "

See chunk 2 of the latest CI run.

walac:
> Hrm, why the switch of this from an arrow function to an old style? Just curious.

`this` is `undefined` with arrow function (don't ask me why).

gregarndt:
That definitely greened up the tests now and the only test failing is one that I expect to fail right now until I look more into it.  Awesome! I think we can look at rolling this out on Monday, what do you think?

walac:
200_s

PR #178, opened 2015-10-29
Bug 1219612 - use task definition from claimTask response

petemoore:
Depends on https://github.com/taskcluster/taskcluster-queue/pull/61 landing first (see [relevant line](https://github.com/taskcluster/taskcluster-queue/pull/61/files#diff-13bae93e5c75b1afda41787e8285918eR34)).

### DIFF
--- a/lib/task_listener.js
+++ b/lib/task_listener.js
@@ -334,7 +334,7 @@ export default class TaskListener extends EventEmitter {
         }
       );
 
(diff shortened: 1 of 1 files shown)


gregarndt:
Looks like tests are green now that the changes to the queue rolled out.  Merging this in, but it won't be deployed immediately for the worker.  Will go out when the next round of changes do.  Thanks for submitting this!

PR #179, opened 2015-11-05
Bug 1220252: use new taskcluster-vpn-proxy image

djmitche:


djmitche:
and yes, it's been pushed :)

PR #180, opened 2015-11-06
Include actual error when extending taskgraph fails

gregarndt:
Scheduler returns a JSON object when there is an error extending a task
graph.  This object appears as [object, object] in the log if nothing
special is done.


gregarndt:
@walac Would you mind taking a look at this?  I also added you to the team org so you should get an invitation for that as well.

PR #181, opened 2015-11-06
Bug 1221491 - Log the feature that could not be started/stopped

gregarndt:
This should now output an error such as:
[taskcluster:error] Task was aborted because states could not be created successfully. Error linking balrogVPNProxy : Insufficient scopes to use 'balrogVPNProxy' feature.  Try adding docker-worker:feature:balrogVPNProxy to the .scopes array.

### DIFF
--- a/lib/features/dind.js
+++ b/lib/features/dind.js
@@ -20,6 +20,7 @@ const INIT_TIMEOUT = 30 * 1000;
 
 export default class DockerInDocker {
   constructor() {
+    this.featureName = 'dind';
     // dind-service container
     this.container = null;
     this.tmpFolder = path.join('/tmp', slugid.v4());
@@ -76,7 +77,7 @@ export default class DockerInDocker {
     };
   }
 
-  killed(task) {
+  async killed(task) {
     task.runtime.gc.removeContainer(this.container.id);
 
     // Remove temporary folder, this should be possible even though container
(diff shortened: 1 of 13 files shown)


gregarndt:
@walac  Could you also take a look at this one?  Thanks!

walac:
Yes, sure.

lib/features/dind.js
```diff
@@ -76,7 +77,7 @@ export default class DockerInDocker {
     };
   }
 
-  killed(task) {
+  async killed(task) {
     task.runtime.gc.removeContainer(this.container.id);
```
walac: A see the `async` keyword but none `await` added, is this fixing some mistake?

gregarndt: This is so the method is treated as a promise for all features so we can call .then() in states.js . Can't call then() on a regular method

lib/task.js
```diff
@@ -702,13 +702,13 @@ export class Task {
...
       catch (e) {
-        debug(e.stack);
         return await this.abortRun(
           'states_failed',
           this.fmtErrorLog(
             'Task was aborted because states could not be started ' +
-            `successfully. ${e.stack || e}`
-          )
+            `successfully. ${e}`
+          ),
```
walac: Isn't this supposed to be `e.message`? (or the code above be simply `e`?)

gregarndt: The line above I think I want to be e.message. This e is passed to abort run and will be logged to paper trail. I want the whole error passed there

PR #185, opened 2015-11-20
Initial hack on noVNC support 

jonasfj:
This is very hard to test automatically... I considered using phantomjs... but I'm not even sure that would work... and it quickly get very very complicated as we need to copy in noVNC sources.

I'm aware that it's not pretty to mix this with the interactive feature. But both of them needs the wrapper script that prevents exit while an interactive session is in progress. Also we don't want two options.


jonasfj:
@gregarndt, I think this is ready for review on Monday...

If you don't have time for a full review, it would be awesome if you could at least build and deploy and AMI as test-worker or something.. .So I can play with the UI side of it... and see how bad the latency is... Maybe I do need to fix latency issues, I'm not sure yet.

jonasfj:
Last two commits can omitted... But I really hope that removing `'interactive.sock'` doesn't break tests..
I explicitly ported the tests to work without... which was the only thing using it...

The idea was good, but websockets doesn't support redirects.. In future we'll add a JSON artifact with the urls for the sockets... at the moment the implication of this is that you can't build custom tools.

jonasfj:
I landed the tools patch without review, but please give it a quick rubberstamp review:
https://github.com/taskcluster/taskcluster-tools/pull/33

deploy/Makefile
```diff
@@ -1,5 +1,5 @@
 CONFIG_FILE?=./deploy.json
-TARGETS?=$(shell find template -type file | sed 's/template/target/')
+TARGETS?=$(shell find template -type f | sed 's/template/target/')
```
gregarndt: thanks for fixing this++

jonasfj: I hope it works on Mac too...

Otherwise we'll have to dig up some old make magic :)

gregarndt: I tested it, it works :)

lib/features/interactive.js
```diff
@@ -9,24 +9,128 @@ import Promise from 'promise';
...
+    AttachStderr: false,
+    AttachStdin: false,
+    Tty: false,
+    Cmd: ['/.taskclusterutils/list-displays']
+  });
+  let stream = await exec.start({
+    Detach: false,
+    stdin:  false,
+    stdout: true,
+    stderr: false,
```
gregarndt: Hrm, will there never be something written to stderr?

jonasfj: Nope... not unless there is a segfault.. .and in that case I just don't care :)

If the output doesn't parse that's just too bad... and we ignore it... This is a best effort service, so we try to fail gracefully..

Note: obviously people can put all sorts of files, folders and domain sockets into `/tmp/.X11-unix/`, and if I try to open those as domain sockets using libX11, I suspect undefined things could happen... So if people intentionally do very obscure things in their container, then absolutely they can build it so that this feature doesn't work.

PR #186, opened 2015-11-25
Bug 1132214 - Allow task definition to explicitly state the location …

JohanLorenzo:
…of the log

### DIFF
--- a/lib/features/local_live_log.js
+++ b/lib/features/local_live_log.js
@@ -8,12 +8,13 @@ import Debug from 'debug';
 import http from 'http';
 import slugid from 'slugid';
 import URL from 'url';
+import path from 'path';
 
 import BulkLog from './bulk_log';
 import waitForPort from '../wait_for_port';
 
-const ARTIFACT_NAME = 'public/logs/live.log';
-const BACKING_ARTIFACT_NAME = 'public/logs/live_backing.log';
+const ARTIFACT_PATH = 'public/logs/live.log';
+const BACKING_ARTIFACT_NAME = 'live_backing.log';
 // Maximum time to wait for the put socket to become available.
 const INIT_TIMEOUT = 2000;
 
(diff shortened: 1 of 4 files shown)


lib/features/local_live_log.js
```diff
@@ -35,6 +35,9 @@ export default class TaskclusterLogs {
     let taskId = task.status.taskId;
     let runId = task.runId;
 
+    let artifactName = task.payload.log || ARTIFACT_NAME;
+    // let artifactName =  || BACKING_ARTIFACT_NAME
+
     // ensure we have a bulk log backing stuff...
     this.bulkLog = new BulkLog(BACKING_ARTIFACT_NAME);
```
gregarndt: Ok, this is also going to need changing.  What happens in the life of the log is:

1. Live log reference artifact is created that points to the container streaming the live log.  This is called a reference/redirect artifact. (this allows someone to view the realtime stream of the log directly from the host.
2. The "bulk log" is created, which is just a temporary file locally first that is the entire log
3. Once the task is complete, the bulk log temporary file is uploaded as an artifact, typically with the name "live_backing.log"
4. the live log redirect artifact is then pointed to that bulk log artifact rather than the container (since the container serving up the live log is going to be gone after the task completes)

If there is a custom name for the live log, such as "/private/jlorenzo/task.log" , then the bulk log must also be stored under this path.  Probably can just do a path.parse(task.payload.log) and toss the backing log under that path when it uploads the artifact.  Hopefully I didn't confuse things.

test/integration/logging/local_live_log_test.js
```diff
@@ -97,4 +97,25 @@ suite('live logging', () => {
...
+          '-c',
+          'echo "first command!"'
+        ],
+        log: CUSTOM_LOCATION,
+        maxRunTime: 3 * 60
+      }
+    });
+    assert.equal(result.run.state, 'completed', 'task should be successful');
+    assert.equal(result.run.reasonResolved, 'completed', 'task should be successful');
+    assert.equal(result.artifacts[CUSTOM_LOCATION], 'foo bar');
```
gregarndt: we probably should make sure that not only the live log is put there, but also the backing log and nothing has been uploaded under the "public/" namespace

gregarndt:
This is great, thank you for hacking on the worker.  I left a couple of comments that I hope will help things.

gregarndt:
I think this can be closed now right as the other PR was where the work took place?

JohanLorenzo:
Absolutely

PR #187, opened 2015-11-26
Update README.md

gregarndt:


README.md
```diff
@@ -73,17 +73,32 @@ for the docker worker in particular these are important:
...
+
+1. `vagrant ssh`
+2. The tests require TASKCLUSTER_ACCESS_TOKEN, TASKCLUSTER_CLIENT_ID, PULSE_USERNAME, PULSE_PASSWORD to be setup within the environment.  If they were not available locally when building, add them to the virtual machine now.
+3. `cd /vagrant` # Your local checkout of the docker-worker repo is made available under the '/vagrant' directory
+4. `./build.sh` # Builds some of the test images that are required
+5. `npm install` # Installs all the necessary node modules
+
+#### Running Tests
+
+1. Either all the tests can be run, but running `npm test` or `./test/test.sh`, however, under most circumstances one only wants to run a single test suite
```
JohanLorenzo: Maybe we can add a paragraph on authentication? I couldn't find out missing scope until I hacked into the TC client to expose the error.

JohanLorenzo:
There's another node v0.12 or greater [here](https://github.com/taskcluster/docker-worker/pull/187/files#diff-04c6e90faac2675aa89e2176d2eec7d8R38)

README.md
```diff
@@ -73,17 +73,33 @@ for the docker worker in particular these are important:
 
 ## Running tests
 
-The `./test/test.sh` script is used to run individual tests which are
-suffixed with `_test.js` for example: `./test/test.sh test/integration/live_log_test.js`.
+There are a few components that must be configured for the tests to work properly (e.g. docker, kernel
```
JohanLorenzo: I just realized these steps are needed if you're on Mac OS X. If you're running Linux, the odds you can't run docker yourself are small.

README.md
```diff
@@ -73,17 +73,33 @@ for the docker worker in particular these are important:
...
+an environment very similar to the one docker-worker runs in production.
 
-```sh
-# from the top level
+#### Setting up vagrant
 
-./build.sh
-npm test
-```
+1. Install virtualbox
```
JohanLorenzo: `[VirtualBox](https://www.virtualbox.org/)`

README.md
```diff
@@ -73,17 +73,33 @@ for the docker worker in particular these are important:
...
+2. The tests require TASKCLUSTER_ACCESS_TOKEN, TASKCLUSTER_CLIENT_ID, PULSE_USERNAME, PULSE_PASSWORD to be setup within the environment.  If they were not available locally when building, add them to the virtual machine now.
+3. `cd /vagrant` # Your local checkout of the docker-worker repo is made available under the '/vagrant' directory
+4. `./build.sh` # Builds some of the test images that are required
+5. `npm install` # Installs all the necessary node modules
+
+#### Running Tests
+
+1. Either all the tests can be run, but running `npm test` or `./test/test.sh`, however, under most circumstances one only wants to run a single test suite
+2. For individual test files, run `./node_modules/mocha/bin/mocha --bail test/<file>`
+3. For running tests within a test file, add "--grep <phrase>" when running the above command to capture just the individual test name.
```
JohanLorenzo: Or you can just make one test run thanks to [test.only()](http://mochajs.org/#exclusive-tests)

README.md
gregarndt: Yea, if one wants to run within docker on their linux host, they are more than capable of doing so.  They will also need some of the kernel modules for video and sound devices that we use if they want the loopback devices to work, which I've seen some people not have configured.

README.md
gregarndt: We absolutely need something like that.  I'm surprised the TC client didn't show the error, it typically does for missing scopes.  If there was an error that wasn't clear I would suggest entering a bug against tc client so that we can make sure we're getting the right errors out there.

JohanLorenzo: Okay, I wasn't sure if that was expected or not. I'll file a bug for that

README.md
gregarndt: fixed

README.md
gregarndt: Great, thanks.  I also added the scopes that were added for your client ID.  I think this list will probably change at some point as we move them into their own role that people could have.

gregarndt:
I think I addressed most of your comments, let me know what you think.

JohanLorenzo:
LGTM

PR #188, opened 2015-11-26
Add guard around starting a docker container

gregarndt:
When a docker container fails to start up either due to a docker issue or bad command, the error was not getting caught properly, resulting the task logs not getting uploaded and task not being resolved until deadline was exceeded or machine was shutdown.


jonasfj:
looks okay to me...

jonasfj:
`r+`

I which we had a better structure where all of these would just be catched... but this is fine..

PR #189, opened 2015-11-27
Bug 1132214 - Allow task definition to explicitly state the location of the log

JohanLorenzo:
Made this PR, in order to reproduce this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1228583

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 # Docker Worker
 
-Docker task host for linux.
+Docker task host for Linux.
 
 Each task is evaluated in an isolated docker container.
 Docker has a bunch of awesome utilities for making this work well...
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 # Docker Worker
 
-Docker task host for Linux.
+Docker task host for linux.
 
 Each task is evaluated in an isolated docker container.
 Docker has a bunch of awesome utilities for making this work well...
--- a/lib/features/local_live_log.js
+++ b/lib/features/local_live_log.js
@@ -8,13 +8,12 @@ import Debug from 'debug';
 import http from 'http';
 import slugid from 'slugid';
 import URL from 'url';
-import path from 'path';
 
 import BulkLog from './bulk_log';
 import waitForPort from '../wait_for_port';
+import getLogsLocationsFromTask from './logs_location.js';
 
-const ARTIFACT_PATH = 'public/logs/live.log';
-const BACKING_ARTIFACT_NAME = 'live_backing.log';
+
 // Maximum time to wait for the put socket to become available.
 const INIT_TIMEOUT = 2000;
 
@@ -30,12 +29,7 @@ export default class TaskclusterLogs {
     this.container = null;
     this.token = slugid.v4();
 
-    this.artifactPath = undefined; // will be overriden in link()
-  }
-
-  static getArtifactPathFromTask(task) {
-    return task && task.payload && task.payload.log ?
-        task.payload.log : ARTIFACT_PATH;
+    this.logsLocations = undefined; // will be overriden in link()
   }
 
   async link(task) {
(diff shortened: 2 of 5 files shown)


JohanLorenzo:
Renamed this PR, because this is the only one (out of 4) that taskcluster-ci picked up

README.md
```diff
@@ -19,7 +19,7 @@
 
 # Docker Worker
 
-Docker task host for linux.
+Docker task host for Linux.
```
JohanLorenzo: Nit: This change is not really necessary anymore.

gregarndt:
Just to note,i do not think that the two failures here are caused by this PR. I've seen it happened elsewhere and I'm not sure why yet but shouldn't hold up this PR.

lib/features/local_live_log.js
```diff
@@ -35,8 +43,13 @@ export default class TaskclusterLogs {
     let taskId = task.status.taskId;
     let runId = task.runId;
 
+    this.artifactPath = TaskclusterLogs.getArtifactPathFromTask(task.task);
+
+    let directory = path.parse(this.artifactPath).dir;
+    let backingPath = directory + '/' + BACKING_ARTIFACT_NAME;
```
gregarndt: This is probably ok, but would look weird if someone named the live log something other than live.log.

For instance if task.payload.log is 'b2g-team/logs/task_log.log' the backing log will be 'b2g-team/logs/live_backing.log'

Sometimes makes me wonder if we either should:
A) parse file name, and add _backing to it
B) make take.payload.log be just a path to put logs under, such as setting it to 'b2g-team/logs'

I'm fine with this for now though as really the only real usecase is for the b2g specific stuff and the tasks can understand that this is what will happen.

test/testworker.js
```diff
@@ -231,7 +231,9 @@ export default class TestWorker extends EventEmitter {
     return await Promise.all(graph.tasks.map(async (task) => {
       // Note: that we assume runId 0 here which is fine locally since we know
       // the number of runs but not safe is we wanted to test reruns.
-      return await this.fetchTaskStats(task.taskId, 0);
+      var LocalLiveLog = require('../lib/features/local_live_log');
```
gregarndt: Do we need to import this here or just import this globally in testworker.js instead of two different locations where it's imported?  Also, what do you think about moving that method out of the class and just have local_live_log.js export that function to be imported independently?  Do you see a strong case for this to be a class method?  I'm impartial either way, just curious.

test/testworker.js
```diff
@@ -156,13 +156,13 @@ export default class TestWorker extends EventEmitter {
   /**
   Fetch all the common stats used by the tests.
   */
-  async fetchTaskStats(taskId, runId) {
+  async fetchTaskStats(taskId, runId, artifactPath) {
```
gregarndt: Rather than passing in the artifactPath, there might be cases in the future where we need other things from the task definition.  Instead of determining the artifact path twice here, what do you think about just passing in the task definition and then we can determine the artifact path here just the one time (instead of for graphs and individual tasks)?

gregarndt:
This looks good, and I don't think the test failures are caused by anything you changed.  My comments are mainly nits, but I do think that maybe we should consider the question I put in testworker.js.

Thanks for helping making this happen.

lib/features/local_live_log.js
JohanLorenzo: Option A makes sense to me. I'll implement it

test/testworker.js
JohanLorenzo: Moving it out sounds fine. I made it as a class method, just to centralize where we define the path. I agree, having it in a smaller class sounds better.

I'll change that

test/testworker.js
JohanLorenzo: Sounds good to me!

test/testworker.js
```diff
@@ -273,6 +277,7 @@ export default class TestWorker extends EventEmitter {
     var runId = status.runs.pop().runId;
 
     // Return uniform stats on the worker run (fetching common useful things).
-    return await this.fetchTaskStats(taskId, runId);
+    task.taskId = taskId;
```
gregarndt: Ok, I found out why the volume cache test was failing after your commits, but passed on master.  This line breaks reusing a task again within the same test.

This "delete task['taskId'];" needs to be added somewhere before posting the second task here:
https://github.com/taskcluster/docker-worker/blob/master/test/integration/container_volume_caching_test.js#L176

JohanLorenzo: Thanks for the information! I'll change that

gregarndt:
I see the new failure pop up, but the one change you made couldn't have caused this.  I'm not entirely sure why this is an issue now, but I'm not going to chalk it up to your changes.

PR #192, opened 2015-12-03
Bug 1221661: add an allowPtrace feature

djmitche:
This is currently only controlled by the feature, meaning it would work in any workerType and only be limited by the feature scope.  I think this is probably fine -- as discussed in the bug, even with this feature enabled, ptrace is limited to processes within the container both by the PID namespace (weak) and the AppArmor policy (strong).

Code is untested but seems to be syntactically correct.

### DIFF
--- a/lib/features/allow_ptrace.js
+++ b/lib/features/allow_ptrace.js
@@ -1,0 +1,25 @@
+/**
+ * This feature adjusts the AppArmor profile to allow ptrace within a container.
+ * All of the interesting action takes place in `lib/task.js`.
+ */
+
+import { scopeMatch } from 'taskcluster-base/utils';
+
+// Prefix used in scope matching for docker-worker features
+const FEATURE_SCOPE_PREFIX = 'docker-worker:feature:';
+
+export default class AllowPtrace {
+  constructor() {
+    this.featureName = 'allowPtrace';
+  }
+
+  async link(task) {
+    let featureScope = FEATURE_SCOPE_PREFIX + this.featureName;
+    if (!scopeMatch(task.task.scopes, [[featureScope]])) {
+      throw new Error(
+        `Insufficient scopes to use '${this.featureName}' feature.  ` +
+        `Try adding ${featureScope} to the .scopes array.`
+      );
+    }
+  }
+}
--- a/lib/task.js
+++ b/lib/task.js
@@ -8,6 +8,8 @@ import uuid from 'uuid';
 import { validator } from 'taskcluster-base';
 import { PassThrough } from 'stream';
 import States from './states';
+import fs from "mz/fs"
+import child_process from "mz/child_process"
 
 import features from './features';
 import getHostname from './util/hostname';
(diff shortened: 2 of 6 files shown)


djmitche:
That second graph has a bunch of failures, but I think they're unrelated to my code.  I'm testing them on a no-change PR now.  I didn't add tests, so that just means I didn't break anything, but hey, that's something!

lib/features/allow_ptrace.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * This feature adjusts the AppArmor profile to allow ptrace within a container.
+ * All of the interesting action takes place in `lib/task.js`.
+ */
+
+export default class AllowPtrace {
```
gregarndt: Wow, this is really highlighting the need for making the "features" betters in the worker. Sorry that you had to do this nasty hack.

lib/task.js
```diff
@@ -273,9 +275,43 @@ export class Task {
         .concat(oldEntrypoint);
     }
 
+    if(this.task.payload.features && this.task.payload.features.allowPtrace) {
```
gregarndt: I think this unfortunately is the easiest least intrusive way of adding this.  In the future as we add more features, I want to create some transformation process for the docker config so it passes the config to each feature handler and it will return a modified config based on whatever it needs to do.

lib/task.js
```diff
@@ -273,9 +275,43 @@ export class Task {
...
+
+    // update the name of the profile
+    profile = profile.replace("profile docker-default", "profile " + name);
+
+    // add a rule allowing ptrace to other processes with this policy, at
+    // the very end
+    profile = profile.replace("\n}", "\n  ptrace (trace,read) peer=" + name + "\n}");
+
+    // write out the new profile
+    await fs.writeFile("/etc/apparmor.d/" + name, profile)
```
gregarndt: nit: we probably don't need to worry about this since instances are not long lived, the profile is small, and the instances do not claim a lot of tasks, but these old apparmor profiles are not cleaned up after the tasks completes. 

Not something that I see as an immediate thing to clean up though.

lib/features/allow_ptrace.js
```diff
@@ -0,0 +1,10 @@
...
+  constructor() {
+    this.featureName = 'allowPtrace';
```
gregarndt: Is there any harm in not limiting this feature by a scope?  Currently all features are allowed unless explicitly set to check a scope in this class.

gregarndt: Also, if this should be something protected behind a scope, we can do what testdroid or balrog proxy do now as the burden of refactoring how the scopes are checked shouldn't be your problem but I would like to get a bug entered so I can move that feature scope checking out of each individual class and have something before that when the features are first initialized.

Perhaps call a method on each feature here to see if the task is authorized to use that feature:
https://github.com/taskcluster/docker-worker/blob/master/lib/task.js#L65

lib/features/allow_ptrace.js
djmitche: https://bugzilla.mozilla.org/show_bug.cgi?id=1229902 might help

lib/features/allow_ptrace.js
djmitche: Hm, I would like to limit this by scope, for sure :)

lib/task.js
djmitche: https://bugzilla.mozilla.org/show_bug.cgi?id=1229902

lib/task.js
djmitche: I should add a comment to that effect, but that was my analysis too

lib/task.js
gregarndt: Ah, great, I'm glad a bug was already entered for that.  I haven't gotten to all my bugmail since PTO so thanks for pointing that one out.

gregarndt: And you're absolutely right, the states as they currently stand do not make a lot of sense for this.  The "feature" handling is a very overloaded thing and needs some refactoring.

djmitche:
OK, fixed those two things.

lib/features/allow_ptrace.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  async link(task) {
+    let docker = task.runtime.docker;
+    let featureScope = FEATURE_SCOPE_PREFIX + this.featureName;
+    if (!scopeMatch(task.task.scopes, [[featureScope]])) {
```
gregarndt: need to import scopeMatch

lib/features/allow_ptrace.js
```diff
@@ -0,0 +1,24 @@
...
+// Prefix used in scope matching for docker-worker features
+const FEATURE_SCOPE_PREFIX = 'docker-worker:feature:';
+
+export default class AllowPtrace {
+  constructor() {
+    this.featureName = 'allowPtrace';
+  }
+
+  async link(task) {
+    let docker = task.runtime.docker;
```
gregarndt: not using docker here.

djmitche: I'm working on writing and running some tests right now, so I'll catch that
kind of thing.

gregarndt: Ok, cool.  Sorry, was just checking some things out while having coffee this morning.  I was just using this basic test: https://gist.github.com/gregarndt/ad2d538307c6d934813d

Also, for the tests to run, I think we need to mount /etc/apparmor.d into the container test/dockerworker.js uses.

lib/task.js
```diff
@@ -273,9 +275,45 @@ export class Task {
...
     return procConfig;
   }
 
+  async makePtraceAppArmorProfile(taskId) {
+    // copy and modify the docker-default profile
+    let profile = await fs.readFile("/etc/apparmor.d/docker");
+    let name = "worker-" + taskId;
+
+    // update the name of the profile
+    profile = profile.replace("profile docker-default", "profile " + name);
```
gregarndt: I believe readFile returns a buffer which will not have the method replace

djmitche:
OK, wrote an integration test that now even passes!  Have another look-see?

djmitche:
I just rebased and pushed an update that:

 * uses `apparmor_parser` instead of restarting the service
 * sets up dockerworker.js to allow use of apparmor within the container (requires mounting securityfs inside the container)
 * passes its tests

gregarndt:
r+, updated bug. Merge if you're happy.

PR #193, opened 2015-12-03
Bug 1229132: implement coalescer

djmitche:
This is still a work in progress.  Remaining:

 * [x] don't explode if a non-primary claim can't be reclaimed (+ test)
 * [x] tests for `TaskListener.coalesceClaim`
 * [x] actually make coalescing calls and interpret the results (+ tests, including for partial and complete claiming failure)
 * [x] add superseded-by artifacts
 * [x] add supersedes artifacts
 * [x] write `fetchCoalescerTasks`


gregarndt:
r+, updated bug.  Merge if you're happy.

djmitche:
OK @gregarndt I think this is ready for your perusal.  Most of the new bits have unit tests.  I don't really want to get the coalescer service involved in integration testing, so I suppose we'll test the whole thing out in ami-test.

lib/task.js
```diff
@@ -440,40 +565,74 @@ export class Task {
...
-  */
-  scheduleReclaim(claim) {
-    // Figure out when to issue the next claim...
-    let takenUntil = (new Date(claim.takenUntil) - new Date());
-    let nextClaim = takenUntil / this.runtime.task.reclaimDivisor;
+   * Resolves all of the non-primary claims for this task, optionally
+   * adding an artifact to each one named "public/superseded-by"
+   * containing the taskId/runId of the primary claim.
+   *
+   * @param {Integer} primaryTaskId taskId of the primary task
```
gregarndt: nit: task ID is not an integer

lib/task.js
```diff
@@ -440,40 +565,74 @@ export class Task {
...
+          });
+
+          supersedes.push(content);
+        }
+      }));
+    } catch (e) {
+      // failing to resolve a non-primary claim is not a big deal: it will
+      // either time out and go back in the queue, or it was cancelled or
+      // otherwise modified while we were working on it.
+      log("while resolving superseded tasks: " + e);
```
gregarndt: Do we care about the primary task ID and the non-primary task ID that it was trying to resolve in this message?

lib/task_listener.js
```diff
@@ -314,11 +319,102 @@ export default class TaskListener extends EventEmitter {
...
+   * Coalesce the task in the given claim, claiming any additional tasks
+   * directly from the TC queue service, then calling runTask with the
+   * resulting task set.
+   */
+  async coalesceClaim(claim) {
+    let task = claim.task;
+    let taskId = claim.status.taskId;
+
+    try {
+      // if the task is not coalescible, then the taskset is just the one claim
```
gregarndt: nit: doesn't need to be fixed now, but instead of reading this and caring how we determine if it's able to be coalesced, we could just have a method that returned a boolean and moved it out of here.

```js
if (!taskIsCoalescible(task)) {
  return [claim];
}
```

lib/task_listener.js
```diff
@@ -314,11 +319,102 @@ export default class TaskListener extends EventEmitter {
...
+      }
+
+      let routes = task.routes;
+      routes = routes.filter(r => r.startsWith(COALESCER_ROUTE_PREFIX));
+      if  (routes.length != 1) {
+        return [claim];
+      }
+
+      let coalescingKey = routes[0].substr(COALESCER_ROUTE_PREFIX.length);
+      let coalescerUrl = task.payload.coalescer.url + coalescingKey;
```
gregarndt: I'm not sure who specifies the coalescer URL and if this is an issue or not, but if the URL does not have a trailing slash, this would result in an invalid URL, right?

schemas/payload.js
```diff
@@ -213,6 +213,22 @@ module.exports = {
         "$ref": "#/definitions/artifact"
       }
     },
+    "coalescer": {
+      "type": "object",
+      "title": "Coalescer information",
+      "desceription": "Information about the coalescer service that should be used to coalesce multiple tasks into a single run.  The worker determines the task's coalescing keys based on task routes with the given prefix, the suffix being the coalescing key.  Each task key is then appeneded to the coalescer URL, which is expected to return a JSON list of taskIds which may be coalesced with this one.  The tasks in the list should be sorted so that tasks earlier in the list supersede (\"should be performed instead of\") tasks later in the list.",
```
gregarndt: nit: 'description'

lib/task_listener.js
```diff
@@ -164,7 +168,8 @@ export default class TaskListener extends EventEmitter {
     if (exceedsThreshold) return;
 
     let claims = await this.taskQueue.claimWork(availableCapacity);
-    claims.forEach(this.runTask.bind(this));
+    let tasksets = await Promise.all(claims.map(this.coalesceClaim.bind(this)));
+    await Promise.all(tasksets.map(this.runTaskset.bind(this)));
```
gregarndt: Wouldn't this block claiming any more work until all calls to runTaskset (meaning the entire task execution) completes?

lib/task_listener.js
```diff
@@ -314,11 +319,102 @@ export default class TaskListener extends EventEmitter {
...
   * Run task that has been claimed.
   */
-  async runTask(claim) {
+  async runTaskset(claims) {
     let runningState;
     try {
+      // the claim we're actually going to execute (the primary claim) is the
+      // first one; the rest will be periodically reclaimed and then resolved,
+      // but will not actually execute.
+      let claim = claims[0];
```
gregarndt: scoping 'claim' here means that it's not available if an error is caught lower down.

lib/task.js
```diff
@@ -440,40 +565,74 @@ export class Task {
...
-        }()
-      }.bind(this), nextClaim);
+        await queue.reportException(taskid, runId, {reason});
+
+        if (addArtifacts) {
+          let task = c.task;
+          // set the artifact expiration to match the task
+          let expiration = task.expires || taskcluster.fromNow(task.deadline, "1 year");
+          let content = {'taskId': primaryTaskId, 'runId': primaryRunId};
+          contentJson = JSON.Stringify(content);
```
gregarndt: the method is 'stringify' (lower case s).

lib/task.js
```diff
@@ -440,40 +565,74 @@ export class Task {
...
+      // failing to resolve a non-primary claim is not a big deal: it will
+      // either time out and go back in the queue, or it was cancelled or
+      // otherwise modified while we were working on it.
+      log("while resolving superseded tasks: " + e);
+    }
+
+    if (addArtifacts) {
+      let task = this.claim.task;
+      let expiration = task.expires || taskcluster.fromNow(task.deadline, "1 year");
+      let contentJson = JSON.Stringify(supersedes)
```
gregarndt: the method is 'stringify' (lower case s).

lib/task.js
```diff
@@ -440,40 +565,74 @@ export class Task {
...
+      await uploadToS3(c, contentJson, "public/supersedes.json", expiration, {
```
gregarndt: 'c' is not defined here.  This is declared in the 'try' block when calling map on this.claims.

gregarndt: In its current form, uploadToS3 takes a task handler as its first argument, and here (if 'c' was a claim), it's just the claim.  Really uploadToS3 should accept a task structure (that is the definition and status, I think those might be in the claim structure?), and a queue instance from the runtime instead or relying so much on the 'task' object that's passed in.

lib/task.js
```diff
@@ -440,40 +565,74 @@ export class Task {
...
-      }.bind(this), nextClaim);
+        await queue.reportException(taskid, runId, {reason});
+
+        if (addArtifacts) {
+          let task = c.task;
+          // set the artifact expiration to match the task
+          let expiration = task.expires || taskcluster.fromNow(task.deadline, "1 year");
+          let content = {'taskId': primaryTaskId, 'runId': primaryRunId};
+          contentJson = JSON.Stringify(content);
+          await uploadToS3(c, contentJson, "public/superseded-by.json", expiration, {
```
gregarndt: see my comment below where uploadToS3 is called.

gregarndt:
Added some comments, some nits, some true bustages.  I debugged this for a bit locally to hopefully give you a head start on identifying the issues.  Hopefully it helps out some.

lib/task.js
djmitche: Heh, makes it pretty obvious I didn't figure out how to unit-test this :)

lib/task.js
djmitche: Actually it takes a claim (requiring `{runId, status: { taskId }, task: { expires }}`)

lib/task_listener.js
djmitche: Oh, good point.  The old code leaves un-handled promises, so I was trying to fix that -- is there a good quick way to say "if this promise fails, log it, otherwise I don't care"?

lib/task_listener.js
djmitche: meh, it's one line, and it's only used here..

lib/task.js
gregarndt: Hrm, then I must be misreading.  Looking at it, it seems that it's taking a complete task handler, because then it looks for task.runtime.queue.
https://github.com/taskcluster/docker-worker/pull/193/files#diff-877d95d813ff49157dff0c2db4fcab51R5

Note: This is going to be redone soon because we do not want to use the global queue client, but rather create a new one using the temp creds given back by a claim. (wander is looking at adding this in q1).

lib/task_listener.js
djmitche: There are arguments on both sides of this: adding the slash is probably user-friendly, but really just papers over a user misconfiguration and is "magic".  The URLs will be in the task definitions, which will be in-tree, so I expect they'll not spontaneously lose their slashes.

lib/task.js
djmitche: Oh, indeed, I missed that.  Hmph.

lib/task_listener.js
gregarndt: In the old code, I think most things in runTask was wrapped in a try/catch and would result in a "task error" event if something was caught.  That was how it was handled before unless there is a particular case where you see an error that wasn't being handled there.

We've done elsewhere somewhere like attaching .catch() on each promise returned and appending the error to a list for it to be logged rather than it throwing an uncaught error or calling reject() on the promise.

Something like this:
https://github.com/taskcluster/docker-worker/blob/49056e24fab16c63979bbad546ccf63fd809ffa1/lib/features/docker_save.js#L92

lib/task.js
djmitche: OK, I put in a refactoring to fix that.  Thanks :)

lib/task_listener.js
djmitche: I did see a case where `this.runtime.stats.record` failed with an un-caught error.  But this probably isn't my issue to fix, so I changed the map to a forEach and thus basically left it as I found it.

djmitche:
The tests aren't running for me locally, but they're doing the same on master, so I'm going to count that as success.

djmitche:
@gregarndt how is this looking?  Should we try it out on ami-test?

gregarndt:
I have not had a chance to go back and look to see why all of the test chunks are now failing with the latest commits.  I can certainly deploy it to an ami if you want to play around with it.

djmitche:
Sure, let's do it :)

lib/features/docker_save.js
```diff
@@ -65,10 +67,11 @@ export default class DockerSave {
       zipStream.on('error', reject);
     });
     let expiration = taskcluster.fromNow(task.runtime.dockerSave.expiration);
+    expiration = new Date(Math.min(expiration, new Date(task.task.expires)));
     let stat = await fs.stat(pathname);
 
-    await uploadToS3(task,
-      fs.createReadStream(pathname),
+    await uploadToS3(task.runtime.queue, task.status.taskId, task.runId,
+                     fs.createReadStream(pathname),
```
gregarndt: any reason for indenting this, but not the other arguments passed in below this line?

lib/task.js
```diff
@@ -440,40 +565,82 @@ export class Task {
...
+        await queue.reportException(taskid, runId, {reason});
+
+        if (addArtifacts) {
+          let task = c.task;
+          // set the artifact expiration to match the task
+          let expiration = task.expires || taskcluster.fromNow(task.deadline, "1 year");
+          expiration = new Date(Math.min(expiration, new Date(task.expires)));
+          let content = {'taskId': primaryTaskId, 'runId': primaryRunId};
+          contentJson = JSON.stringify(content);
+          await uploadToS3(queue, taskid, runId, contentJson,
```
gregarndt: should be 'taskId' (capital 'I' in 'Id')

lib/task.js
```diff
@@ -440,40 +565,82 @@ export class Task {
...
-          await this.reclaimTask();
-        }()
-      }.bind(this), nextClaim);
+        await queue.reportException(taskid, runId, {reason});
+
+        if (addArtifacts) {
+          let task = c.task;
+          // set the artifact expiration to match the task
+          let expiration = task.expires || taskcluster.fromNow(task.deadline, "1 year");
+          expiration = new Date(Math.min(expiration, new Date(task.expires)));
```
gregarndt: I troubleshooted this a little bit, and it looks like this results in "invalidate date" because expiration here is a string, and when calling Math.min() involving a string results in NaN.  pass that to Date() and things are not so good..

> new Date(Math.min("2016-12-22T18:04:45.420Z", new Date()))
Invalid Date

djmitche: Now that I look at it, this isn't even necessary in this case (it was more or less copy/pasted)

lib/task.js
```diff
@@ -151,20 +153,133 @@ function buildDeviceBindings(devices, taskScopes) {
...
+      // claim
+      this.claim.task = task;
+    } catch (e) {
+      let errorMessage = `Could not reclaim task. ${e.stack || e}`;
+      this.log('error reclaiming task', {error: errorMessage});
+
+      // If this is not the primary claim, just stop trying to reclaim.  The task
+      // will attempt to resolve it as superseded, and fail, but the primary task
+      // and the other superseded tasks will still be resolved correctly.
+      if (this.claim != this.primaryClaim) {
```
gregarndt: So here this doesn't evaluate to true when there is only one task.  I believe perhaps it could be because both are objects, are not the same objects in memory, so they do not evaluate to true.

lib/task.js
```diff
@@ -151,20 +153,133 @@ function buildDeviceBindings(devices, taskScopes) {
...
+      // claim
+      this.claim.task = task;
+    } catch (e) {
+      let errorMessage = `Could not reclaim task. ${e.stack || e}`;
+      this.log('error reclaiming task', {error: errorMessage});
+
+      // If this is not the primary claim, just stop trying to reclaim.  The task
+      // will attempt to resolve it as superseded, and fail, but the primary task
+      // and the other superseded tasks will still be resolved correctly.
+      if (this.claim.status.taskid != this.primaryClaim.status.taskId) {
```
gregarndt: this.claim.status.taskId (capital 'I' in 'Id') instead and the test should be green.

djmitche: d'oh

test/reclaimer_test.js
```diff
@@ -0,0 +1,128 @@
...
+  var Reclaimer = require('../lib/task').Reclaimer;
+  var fakeRuntime, fakeTask;
+  var reclaims;
+  var taskAction;
+  var soon;
+  var reclaimer;
+  var fakeLog = require('debug')('fakeRuntime.log');
+
+  setup(function() {
+    reclaims = [];
```
gregarndt: Do we want to set this back to an empty array after each test?

gregarndt: Sorry ignore this, I misread and thought this was not happening in setup().

test/reclaimer_test.js
```diff
@@ -0,0 +1,128 @@
...
+
+    fakeRuntime.queue.reclaimTask = async function(taskId, runId) {
+      var err = new Error("uhoh");
+      err.statusCode = 409;
+      throw err;
+    };
+
+    await reclaimer.reclaimTask();
+    assert.deepEqual(reclaims, []);
+    assert.equal(taskAction, null);
```
gregarndt: This assertion is failing because cancel is called

djmitche: Fixed, thanks
​

gregarndt:
This last graph looks pretty green to me other than the expected failure I see on chunk 4 (and also in the 'all' task).  If you're good with this now (I am), then I can merge and deploy it to a worker of your choosing to play around with if you want.

PR #194, opened 2015-12-06
Update package.json

sheerun:
- This package is not published anywhere so I mark it as private
- index.js is not existent, so I remove it
- Update urls for repository / issues / bugs


gregarndt:
Thanks for taking the time to look at this!

PR #195, opened 2015-12-07
Make worker vagrant environment more general purpose for users

gregarndt:
Just some minor changes to the readme and vagrant files to make it clearer what is getting installed and that this environment is available.  Easier to point people at the vagrant.sh file now for configuring devices locally when they can't use the vagrant setup.


jonasfj:
Nice,  from just a quick look...

PR #199, opened 2015-12-30
add better logging for coalescing, use routePrefix

djmitche:
Hopefully this will tell me why, when things are not coalesced.


djmitche:
OK, this seems to be working in ami-test right now.

gregarndt:
Looks good to me and looking at the tests it didn't break anything that wasn't broken before.

gregarndt:
Let me know what button pushing you would like.  I can merge and deploy it wherever you want.

djmitche:
Do you want to get it merged and amis generated?  I expect my tree is not up to date with master, so the AMIs I generated are probably not good to deploy.

Once that's in place, shoot me the ami IDs in email and I can start plugging them into the appropriate workerTypes one at a time.

PR #202, opened 2016-01-11
Bug 1237663: use a newer kernel

djmitche:
The new source_ami picks the AMI built on the same data as the HVM source AMI.

I removed `-v` from `set` because it just duplicates every command, which gets pretty confusing.

I'll hold off landing this until I've seen that it makes the `m1.medium` instances not explode, and doesn't break the HVM instances.

### DIFF
--- a/deploy/packer/base/scripts/packages.sh
+++ b/deploy/packer/base/scripts/packages.sh
@@ -80,7 +80,7 @@ sudo apt-get update -y
 }
 
 ## upgrade the kernel, along with extra (which adds AUFS support)
-KERNEL_VER=3.19.0-43-generic
+KERNEL_VER=4.2.0-23-generic
 sudo apt-get install -y \
     linux-image-${KERNEL_VER} \
     linux-headers-${KERNEL_VER} \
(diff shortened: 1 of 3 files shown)


deploy/packer/base/scripts/packages.sh
```diff
@@ -13,17 +13,38 @@ sudo usermod -a -G docker $USER
   apt-get install apt-transport-https
 }
 
+# For reasons that aren't at all clear, packer obliterates sources.list for paravirtualized AMIs.
+# So we put it back with apt-get clean
+sudo apt-get clean
+sudo apt-get update -y
```
gregarndt: There is also the possibility of apt-get update running on line 12.  Could this block just be moved up before line 11 and then remove line 12?

gregarndt:
:+1: Looks good to me, just left a comment, which might be just for my own clarification.

deploy/packer/base/scripts/packages.sh
djmitche: I like that, yes.

djmitche:
Another variant of this got landed elsewise -- this is no longer necessary or even correct.

PR #203, opened 2016-01-11
Bug 1220738: Use temp credentials on task behalf. r=garndt

walac:
For all task related queue operations, we use a queue object created
with temporary credentials.

### DIFF
--- a/lib/features/taskcluster_proxy.js
+++ b/lib/features/taskcluster_proxy.js
@@ -4,6 +4,7 @@ allows tasks to talk directly to taskcluster services over a http proxy which
 grants a particular permission level based on the task scopes.
 */
 import waitForPort from '../wait_for_port';
+import http from 'http';
 
 // Alias used to link the proxy.
 const ALIAS = 'taskcluster';
@@ -28,21 +29,17 @@ export default class TaskclusterProxy {
     var image = task.runtime.taskclusterProxyImage;
     var imageId = await task.runtime.imageManager.ensureImage(image, process.stdout);
 
-    var assumedScope = 'assume:worker-id:' + task.runtime.workerGroup +
-                       '/' + task.runtime.workerId;
-
     var cmd = [
-        '--client-id=' + task.runtime.taskcluster.clientId,
-        '--access-token=' + task.runtime.taskcluster.accessToken
+        '--client-id=' + task.claim.credentials.clientId,
+        '--access-token=' + task.claim.credentials.accessToken
     ];
     // only pass in a certificate if one has been set
-    if (task.runtime.taskcluster.certificate) {
-        cmd.push('--certificate=' + task.runtime.taskcluster.certificate);
+    if (task.claim.credentials.certificate) {
+        cmd.push('--certificate=' + task.claim.credentials.certificate);
     }
-    cmd.push(
-        task.status.taskId,
-        assumedScope
-    );
+    const cert = JSON.parse(task.claim.credentials.certificate);
+    cmd.push(task.status.taskId);
+    cmd = cmd.concat(cert.scopes);
 
     // create the container.
     this.container = await docker.createContainer({
@@ -83,6 +80,42 @@ export default class TaskclusterProxy {
     } catch (e) {
       throw new Error('Failed to initialize taskcluster proxy service.');
     }
+
+    // Update credentials in proxy
+    task.on('credentials', async (credentials) => {
+      let credentials = JSON.stringify(task.claim.credentials);
+
+      await new Promise((accept, reject) => {
+        let req = http.request({
+          hostname: inspect.NetworkSettings.IPAddress,
+          method: 'PUT',
+          path: '/credentials',
+          headers: {
+            'Content-Type': 'text/json',
+            'Content-Length': credentials.length
+          }
+        }, res => {
+          if (res.statusCode == 200) {
+            task.runtime.log('Credentials updated', {
+              taskId: task.status.taskId,
+              runId: task.runId
+            });
+            accept();
+          } else {
+            let err = new Error(`Credentials update failed ${res.statusCode}`);
+            task.runtime.log(err, {
+              taskId: task.status.taskId,
+              runId: task.runId
+            });
+            reject(err);
+          }
+        });
+
+        req.on('error', err => reject(err));
+        req.write(credentials);
+        req.end();
+      });
+    });
 
     return {
       links: [{name, alias: ALIAS}],
--- a/lib/features/taskcluster_proxy.js
+++ b/lib/features/taskcluster_proxy.js
@@ -27,7 +27,7 @@ export default class TaskclusterProxy {
 
     // Image name for the proxy container.
     var image = task.runtime.taskclusterProxyImage;
-    var imageId = await task.runtime.imageManager.ensureImage(image, process.stdout);
+    var imageId = await task.runtime.imageManager.ensureImage(image, process.stdout, task);
 
     var cmd = [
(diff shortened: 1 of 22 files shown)


lib/task.js
```diff
@@ -283,6 +286,8 @@ export class Task {
     this.taskState = 'pending';
     this.options = options;
 
+    this.queue = this.createQueue(this.claim.credentials, runtime);
```
gregarndt: hrm, should this be this.task.queue?

gregarndt: Ignore this, because we have a Reclaimer object per task to claim, the queue instance we use should be for that specific claim, so not on task.queue itself.

lib/task.js
```diff
@@ -221,11 +223,12 @@ export class Reclaimer {
     this.log('reclaiming task');
 
     try {
-      this.claim = await this.runtime.queue.reclaimTask(
+      this.claim = await this.task.queue.reclaimTask(
```
gregarndt: We should have a queue client per Reclaimer because the creds will be different per task we're claiming.  This happens when there are coalesced tasks where there is the primary task we're executing, but we need to reclaim and resolve each individual task that was coalesced.

lib/task.js
```diff
@@ -221,11 +223,12 @@ export class Reclaimer {
...
                                 this.claim.status.taskId, this.claim.runId);
       // reclaim does not return the task, so carry that forward from the previous
       // claim
       this.claim.task = task;
+      this.task.queue = this.task.createQueue(this.claim.credentials, this.runtime);
```
gregarndt: see comment about creating a queue client per reclaimed task.

lib/task.js
```diff
@@ -967,4 +972,29 @@ export class Task {
...
+
+  /**
+  Create a new queue using temp credentials.
+
+  @param {credentials} Temporary credentials.
+  @param {runtime} Runtime object.
+
+  @return New queue.
+  */
+  createQueue(credentials, runtime) {
```
gregarndt: I dont' think this should be a method on the task handler itself.

lib/task.js
```diff
@@ -967,4 +972,29 @@ export class Task {
...
+    return new taskcluster.Queue({
+      credentials: credentials,
+      stats: base.stats.createAPIClientStatsHandler({
```
gregarndt: Because we need a queue client per claim we're doing (coalesced tasks), I'm not sure if we really need stats from the secondary claims it's making, just for the primary one should be good.

lib/task.js
walac: This is to fake createQueue in the reclaimer_test.

lib/task.js
```diff
@@ -221,11 +222,16 @@ export class Reclaimer {
     this.log('reclaiming task');
 
     try {
-      this.claim = await this.runtime.queue.reclaimTask(
-                                this.claim.status.taskId, this.claim.runId);
+      let queue = this.task.createQueue(this.claim.credentials);
```
gregarndt: I think that we only would care about stats for the primary claim we're doing

lib/task.js
```diff
@@ -283,6 +289,8 @@ export class Task {
     this.taskState = 'pending';
     this.options = options;
 
+    this.queue = this.createQueue(this.claim.credentials, runtime);
```
gregarndt: I don't think this is necessary yet, no need to keep a queue client around for each reclaimer...jsut recreate it when it's ready to be reclaimed.

lib/task.js
```diff
@@ -221,11 +222,16 @@ export class Reclaimer {
...
-      this.claim = await this.runtime.queue.reclaimTask(
-                                this.claim.status.taskId, this.claim.runId);
+      let queue = this.task.createQueue(this.claim.credentials);
+      this.claim = await queue.reclaimTask(
+                    this.claim.status.taskId, this.claim.runId);
       // reclaim does not return the task, so carry that forward from the previous
       // claim
       this.claim.task = task;
+
+      if (this.claim.status.taskId == this.primaryClaim.status.taskId) {
```
gregarndt: I would have this replace line 292.  The only time we want to create it is when it's the primary task so that it can be used by the artifact handlers.

lib/task.js
```diff
@@ -967,4 +975,15 @@ export class Task {
...
   }
+
+  /**
+  Create a new queue using temp credentials.
+
+  @param {credentials} Temporary credentials.
+
+  @return New queue.
+  */
+  createQueue(credentials) {
```
gregarndt: Perhaps the status stuff is just added back here.  I'm not sure if we need them, or if we don't. Safer to leave them and sort it out later.

walac:
Depends on https://github.com/taskcluster/taskcluster-auth/pull/36

lib/features/taskcluster_proxy.js
```diff
@@ -84,6 +85,34 @@ export default class TaskclusterProxy {
...
+          hostname: inspect.NetworkSettings.IPAddress,
+          method: 'PUT',
+          path: '/credentials',
+          headers: {
+            'Content-Type': 'text/json',
+            'Content-Length': credentials.length
+          }
+        }, res => {
+          if (res.statusCode == 200) {
+            task.runtime.log('Credentials updated');
```
gregarndt: can we include the task information (id, run #)

lib/features/taskcluster_proxy.js
```diff
@@ -84,6 +85,34 @@ export default class TaskclusterProxy {
...
+            accept();
+          } else {
+            reject(new Error(`Credentials update failed ${res.statusCode}`));
```
gregarndt: before rejecting, could we log this failure like we do above for when it's successful?

test/integration/reclaim_test.js
```diff
@@ -87,5 +88,39 @@ suite('Reclaiming task', function() {
...
+  test('test taskcluster-proxy credentials update', async function() {
+    let payload = {
+      storageType: 'reference',
+      expires: expires().toJSON(),
+      contentType: 'text/html',
+      url: 'https://mozilla.com'
+    };
+
+    let updated = false;
+    worker.on('Credentials updated', function() {
```
gregarndt: This gets called? if I recall this only gets emitted when the task is reclaimed.  Are these tasks getting reclaimed so fast that this test passes?

walac: [This](https://github.com/taskcluster/docker-worker/blob/master/test/integration/reclaim_test.js#L22) guarantees reclaim will happen.

gregarndt: Ah, so yea, the tasks are getting reclaimed at a very fast rate.  Thanks!

gregarndt:
This is looking good however I would like to deploy a worker and push some graphs or something to it to see how it does with some of our production jobs, just to make sure things are covered.  We can't do that until the auth changes land.

gregarndt:
Also I would like to rerun this graph once the auth changes are fixed just to make sure the errors go away.  There was one error for the reclaim test that I woudln't have expected to fail right now but it is.

gregarndt:
It ocurred to me that we probably need to change this as well. https://github.com/taskcluster/docker-worker/blob/master/lib/docker/artifact_image.js#L34

walac:
> ocurred to me that we probably need to change this as well. https://github.com/taskcluster/docker-worker/blob/master/lib/docker/artifact_image.js#L34

@gregarndt Added a new commit to fix it

gregarndt:
I think this is looking good from the quick look I gave it.  I would like to see this passing once the auth stuff is merged before we do more with this.  Hopefully that will be soon.

PR #204, opened 2016-01-12
Bug 1239011: Rename "isolatedcontainers" term to "restrictCPU". r=garndt

walac:


config/defaults.js
```diff
@@ -65,9 +65,9 @@ module.exports = {
   // Run test only teardown and logging events.
   testMode: false,
 
-  // Run each container in as isolated fashion as possible (one core per container)
+  // Run each container in as retrict fashion as possible (one core per container)
```
gregarndt: restrict

PR #205, opened 2016-01-13
Bug 1237663: build PV instances

djmitche:
The key here is the source AMI for PV.  ami-ed051f8c matches the hvmSourceAMI exactly, down to the datestamp (20150315).  Using a newer AMI, even after upgrading docker and the kernel, causes failures -- see bug 1237663 for details.

This change will have no effect for HVM AMIs.


gregarndt:
lgtm

PR #206, opened 2016-01-21
Bug 1234929 - pass certificate to taskcluster-proxy if it is set

petemoore:
Note, I'm a javascript/node.js newbie, so the syntax might be suboptimal!

### DIFF
--- a/lib/features/taskcluster_proxy.js
+++ b/lib/features/taskcluster_proxy.js
@@ -36,13 +36,13 @@ export default class TaskclusterProxy {
         '--access-token=' + task.runtime.taskcluster.accessToken
     ];
     // only pass in a certificate if one has been set
-    if (task.runtime.taskcluster.certificate && task.runtime.taskcluster.certificate.length > 0) {
-        cmd = cmd.concat('--certificate=' + task.runtime.taskcluster.certificate);
+    if (task.runtime.taskcluster.certificate) {
+        cmd.push('--certificate=' + task.runtime.taskcluster.certificate);
     }
-    cmd = cmd.concat([
+    cmd.push(
         task.status.taskId,
         assumedScope
-    ]);
+    );
 
     // create the container.
     this.container = await docker.createContainer({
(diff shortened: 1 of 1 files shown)


lib/features/taskcluster_proxy.js
```diff
@@ -33,10 +33,16 @@ export default class TaskclusterProxy {
 
     var cmd = [
         '--client-id=' + task.runtime.taskcluster.clientId,
-        '--access-token=' + task.runtime.taskcluster.accessToken,
+        '--access-token=' + task.runtime.taskcluster.accessToken
+    ];
+    // only pass in a certificate if one has been set
+    if (task.runtime.taskcluster.certificate && task.runtime.taskcluster.certificate.length > 0) {
```
gregarndt: I don't think you need ot check for length because if the string is empty will fail the check for it being defined.

lib/features/taskcluster_proxy.js
```diff
@@ -33,10 +33,16 @@ export default class TaskclusterProxy {
...
+        cmd = cmd.concat('--certificate=' + task.runtime.taskcluster.certificate);
```
gregarndt: can use cmd.push() instead as I think concat creates a new array and returns it, where as you could just callpush and it will append it to the existing array.

lib/features/taskcluster_proxy.js
```diff
@@ -33,10 +33,16 @@ export default class TaskclusterProxy {
...
+    }
+    cmd = cmd.concat([
```
gregarndt: same as my comment above.

gregarndt:
The two failures I see on the task graph is caused by a known issue with killing a container (only happens in a certain sequence of tests executed) and the other is by a change int he messaging for invalid scopes on the queue/scheduler side of things.  That test needs to be updated and wcosta is aware of it.

gregarndt:
Failures here are not related to the changes.  Thanks!

petemoore:
Thanks!

PR #207, opened 2016-01-21
Update taskgraph test. r=garndt

walac:
Taskgraph error for missing scopes has changed, we then update the test
case accordingly.


petemoore:
Thanks!

PR #208, opened 2016-01-26
bug 1241624 failed tasks should not commit to Balrog

nhirata:


walac:
Do this happen today? The action is actually controlled by mozharness, so am not completely sure this will fix it. Have you tested?

gregarndt:
this is being discussed in the bug, this is not the right place to accomplish what they're looking for.  REally it's something in mozharness (like you said) that should control it

gregarndt:
Since this is being discussed in the bug and that this is not the responsibility of the worker, I'm going to close this pull request.

PR #209, opened 2016-01-28
Bug 1237300: simplify coalescing to superseding

djmitche:
* rename /coalesc(er|ed|ing|ible)/ appropriately
* change the payload format to use only `supersedingUrl`
* append `?taskId=<taskId>` to the superseding URL
* expect an object with property `supersedes`
* require that the resulting list contain the initial taskId

@dividehex let me know if this is missing something


lib/task_listener.js
```diff
@@ -322,43 +322,40 @@ export default class TaskListener extends EventEmitter {
...
-      let coalescingKey = routes[0].substr(routePrefix.length);
-      let coalescerUrl = task.payload.coalescer.url + coalescingKey;
-
-      let tasks = (await this.fetchCoalescerTasks(coalescerUrl))[coalescingKey];
-      if (tasks.length == 0) {
-        this.runtime.log('not coalescing', {taskId, coalescerUrl,
-                           message: 'no tasks to coalese with'});
+      // if the returned list does not contain the initial taskId, then ignore
+      // the request; this is an invalid response from the superseder
+      let taskIdIndex = tasks.indexOf(taskId);
```
gregarndt: if you don't care about the array position, you could do:
`if (!tasks.includes(taskId)) {`

djmitche: Good point -- in an earlier revision, I did care, but now I don't :)

lib/task_listener.js
```diff
@@ -322,43 +322,40 @@ export default class TaskListener extends EventEmitter {
...
 
-      let routes = task.routes;
-      let routePrefix = task.payload.coalescer.routePrefix + ".";
-      routes = routes.filter(r => r.startsWith(routePrefix));
-      if  (routes.length != 1) {
-        this.runtime.log('not coalescing', {taskId, message: 'no single coalescer route in payload'});
+      let supersederUrl = task.payload.supersederUrl;
+
+      let tasks = (await this.fetchSupersedingTasks(supersederUrl, taskId));
+      if (!tasks || tasks.length == 0 || tasks[0] === taskId) {
```
gregarndt: I think I might be misreading something here, but I was looking at `tasks[0] === taskId` and while 
looking over the documentation here [1] it seems that you could have a response such that task[0] is the task claimed initially, but there might be task[1] and task[2] where task[2] should be the the task executed (if it can be claimed).  Should this be `tasks[tasks.length -1] === taskId` ?

[1] https://github.com/taskcluster/taskcluster-docs/pull/69/files#diff-dc8e655f527336a70d0dea380d8ec57aR405

test/superseding_test.js
```diff
@@ -0,0 +1,130 @@
...
+    .reply(200, {'supersedes': ['cTask1', 'cTask2']});
+    var task = makeTask(SUPERSEDER_URL);
+    var claim = makeClaim('fakeTask', 0, task);
+    claimTaskResponses['cTask1'] = {status: {taskId: 'cTask1'}, runId: 0}
+    assert.deepEqual(await listener.applySuperseding(claim), [claim]);
+  });
+
+  test("when the superseder times out, just yield the original claim",
+       async function() {
+    listener.coalescerTimeout = 100;
```
gregarndt: should this be supersederTimeout?

test/superseding_test.js
```diff
@@ -0,0 +1,130 @@
...
+
+  var makeClaim = function(taskId, runId, task) {
+    return {
+      status: { taskId: taskId },
+      runId,
+      task,
+    }
+  };
+
+  var expectNoCoalescerFetch = function() {
```
gregarndt: Do we want to abolish the word "coalescer" from everywhere?

lib/task_listener.js
djmitche: Actually the check shouldn't be there at all.  I put it in when I was thinking we would only claim until we hit the initial taskId -- then I realized that was wrong, and missed changing this bit.  You're a very thorough reviewer!

djmitche:
@gregarndt I built an AMI with this to test, and then forgot it never landed -- is there anything else blocking a landing here?

gregarndt:
I don't believe so!

djmitche:
OK, merged.  @dividehex FYI this will have been rolled back today on desktop-test*, where we just deployed a new AMI.  It should still be in place on the builder workerTypes.

The next deployment should have this patch again.

PR #212, opened 2016-02-18
Apply security updates on boot

gregarndt:
This should run the unattended upgrade procedure during the cloud init phase.

Tested it out with an ami that was create with this and without.  glibc appears to be updated on the one with these statements 

before:
ubuntu@ip-172-31-40-141:~$ ldd --version
ldd (Ubuntu EGLIBC 2.19-0ubuntu6.6) 2.19

after:
ubuntu@ip-172-31-23-40:~$ ldd --version
ldd (Ubuntu EGLIBC 2.19-0ubuntu6.7) 2.19

### DIFF
--- a/deploy/packer/app.json
+++ b/deploy/packer/app.json
@@ -4,8 +4,8 @@
   "variables": {
     "npmPackage":          "",
     "templateContents":    "",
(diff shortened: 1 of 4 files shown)


selenamarie:
:+1:

PR #213, opened 2016-02-22
Bug 1250316 - Compare downloaded file size for indexed image downloads

gregarndt:
There are times where the download will be ended but the entire artifact was now downloaded.  When this happens, the sizes should be compared, and an exception thrown if they do not match.


gregarndt:
While I was able to green up a couple of the tests with this patch, there is one lingering bug in one of the tests that's not related to changes here.  Will need to address that in a future PR.

djmitche:
Looks good

PR #215, opened 2016-03-10
Always precede [taskcluster:error] with newline

djmitche:
This ensures that, even if the previous output wasn't
newline-terminated, the error message appears at the beginning of the
line where treeherder expects to find it.


gregarndt:
Looks good, I'll merge it but it'll probably be early next week (Monday) when this rolls out in a new ami if that's ok.

djmitche:
I think that timing is fine.

PR #217, opened 2016-03-16
Add timings for image download and loading

gregarndt:
A considerable amount of time can be spent downloading and loading image.  This will allow us to report these metrics for downloading, loading (this is specific to task images), and the complete process.

### DIFF
--- a/lib/docker/artifact_image.js
+++ b/lib/docker/artifact_image.js
@@ -70,31 +70,36 @@ export default class ArtifactImage {
 
     let originalTarball = path.join(downloadDir, 'image.tar');
 
-    let newTarball;
     try {
-      await downloadArtifact(
-        this.queue,
-        this.stream,
-        this.taskId,
-        this.artifactPath,
-        originalTarball,
-        this.runtime.dockerConfig
+      await this.runtime.stats.timeGen('taskImageDownloadTime',
+        downloadArtifact(
+          this.queue,
+          this.stream,
+          this.taskId,
+          this.artifactPath,
+          originalTarball,
+          this.runtime.dockerConfig
+        )
       );
-
     } catch(e) {
       await removeDir(downloadDir);
       debug(`Error loading docker image. ${e.stack}`);
       throw new Error(`Error loading docker image. ${e.message}`);
     }
 
+    await this.runtime.stats.timeGen('taskImageLoadTime',
+        this.renameAndLoad(this.imageName, originalTarball)
+    );
+  }
+
+  async renameAndLoad(imageName, originalTarball) {
     this.stream.write(fmtLog('Loading docker image from downloaded archive.'));
     debug('Renaming image and creating new archive');
-    newTarball = await this.renameImageInTarball(this.imageName, originalTarball);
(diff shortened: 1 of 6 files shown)


gregarndt:
@djmitche  mind giving this a look over.  This change is related to just storing the data within influx.  Fixing log messages so others can parse the logs for the data would be done under a different PR.

lib/docker/artifact_image.js
```diff
@@ -70,32 +70,37 @@ export default class ArtifactImage {
...
-    let newTarball;
     try {
-      await downloadArtifact(
-        this.queue,
-        this.stream,
-        this.taskId,
-        this.artifactPath,
-        originalTarball,
-        this.runtime.dockerConfig
+      await this.runtime.stats.timeGen('taskImageDownloadTime',
```
djmitche: (really minor) I kind of feel like this is an abuse of Promises -- in a sync context, this would be `timeGen(() => downloadArtifact(..))`, making it clear that timeGen is responsible for the whole execution of downloadArtifact.  Here, timeGen is taking a Promise on the assumption that it was just begun, and timing it until resolution.

Totally not worth fixing -- I just find the interface weird.

test/garbage_collection_test.js
```diff
@@ -26,7 +26,12 @@ suite('garbage collection tests', function () {
       delayFactor: 15 * 1000,
       randomizationFactor: 0.25
     },
-    log: logger()
+    log: logger(),
+    stats: {
+      timeGen: async function (series, fn) {
+        return await fn;
```
djmitche: Would it be helpful to enforce that `fn` is a Promise (not a function..) here?  If it's not a Promise, then the timing will be 0.

djmitche:
That's all pretty minor -- otherwise, this looks good

lib/docker/artifact_image.js
gregarndt: That's a completely fair point; I agree!.  It's something that's been done in docker-worker for awhile now, and not for a really good reason.

This wouldn't be something that we would want to continue doing in new workers.

test/garbage_collection_test.js
gregarndt: We're not recording timings (and don't care about them either) here in the test.  This is mainly just so that when we use timeGen elsewhere in the image manager stuff it doesn't puke saying that timeGen is not a function when it attempts to use it.

djmitche: Right, this would just be a way to catch cases where timeGen is handed a second argument that is not a Promise -- `timeGen('some-metric', someFunction(..))` looks legitimate, but if someFunction isn't async, some-metric will always be zero even if someFunction is slow.

gregarndt: Would checking fn.then == function be enough to assert that it's a promise in this case?

djmitche: Sounds good, yeah.  I try to add checks like this to test mocks just to avoid subtle errors.

gregarndt: I believe ea2b748 should address this issue.

djmitche:
:thumbsup:

PR #218, opened 2016-03-16
Include timestamp for worker specific log messages

gregarndt:
Output would look like this:
https://tools.taskcluster.net/task-inspector/#7bfM6qMkSI-2B1xs5JvRcA/0

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -18,7 +18,8 @@ export function createLogger(details) {
 
 export function fmtLog() {
   let args = Array.prototype.slice.call(arguments);
-  return '[taskcluster] ' + util.format.apply(this, args) + '\r\n';
+  let ts = new Date().toISOString().replace('T', ' ');
+  return '[taskcluster ' + ts + '] ' + util.format.apply(this, args) + '\r\n';
 }
 
 export function fmtErrorLog() {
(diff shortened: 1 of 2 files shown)


gregarndt:
@djmitche What do you think about adding these time stamps to the [taskcluster] prefixed lines?  Think we could ship it without upsetting people?

lib/log.js
```diff
@@ -18,12 +18,21 @@ export function createLogger(details) {
...
+ * it is at the beginning of the line where treeherder expects it, even if
+ * the last output was not newline-terminated.
+ */
 export function fmtErrorLog() {
   let args = Array.prototype.slice.call(arguments);
-  // always include a newline before this string, so that it is at the beginning of the line
-  // where treeherder expects it, even if the last output was not newline-terminated
-  return '\r\n[taskcluster:error] ' + util.format.apply(this, args) + '\r\n';
+  let msg = fmtLog(...args);
+  return msg + '[taskcluster:error] ' + util.format.apply(this, args) + '\r\n';
```
djmitche: I kinda don't think this duplication is worthwhile -- I suspect we don't want statistics on how long it takes for error XYZ to occur, so I don't think the timestamp is especially useful here.  If I'm wrong and it is useful, then this is a good approach.

gregarndt: Yea that's a good point.  I reverted that part of it and will only include timestamps on the non-error messages.

gregarndt:
If the removal of that extra timestamped line is good for you, I'll merge.

djmitche:
Looks good

PR #219, opened 2016-03-20
[WIP] Add debugger integration (DO NOT MERGE)

jonasfj:
This should be all it takes... once the debugger is written...

The debugger isn't written so let's not roll this just yet :)


gregarndt:
I think we can close this for now, been in a "DO NOT MERGE" state since March.

PR #222, opened 2016-04-12
Accept gzipped responses

gregarndt:


lib/util/artifact_download.js
```diff
@@ -38,8 +38,11 @@ export default async function(queue, stream, taskId, artifactPath, destination,
     try {
       let expectedSize = 0;
       let receivedSize;
-      let startTime = Date.now();
```
gregarndt: Lingering variable that's not used.

PR #225, opened 2016-04-27
Shrinkwrap

imbstack:
https://bugzilla.mozilla.org/show_bug.cgi?id=1259203

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,9 +67,7 @@
     "ws": "^0.8.0",
     "memory-streams": "^0.1.0",
     "dev-null": "^0.1.1",
-    "express": "^4.9.0"
-  },
-  "devDependencies": {
+    "express": "^4.9.0",
     "aws-sdk": "^2.1.5",
     "aws-sdk-promise": "0.0.2",
     "cli-color": "^0.3.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,9 +67,7 @@
     "ws": "^0.8.0",
     "memory-streams": "^0.1.0",
     "dev-null": "^0.1.1",
-    "express": "^4.9.0"
-  },
-  "devDependencies": {
+    "express": "^4.9.0",
```
gregarndt: With shrinkwrap, are we not allowed to have dev dependencies any longer?  Not that it's a huge deal, but it appears that the dev dependencies are mainly there for deploying the worker rather than running it.

imbstack: shrinkwrap in node < 5 is broken w.r.t. devDependencies. We're just avoiding using them until a project is on node 5.

gregarndt:
Looks like #223 was opened for this same thing, but the shrinkwrap has less lines in it even though the commits might be changed.  Perhaps we could get Anup to do it again, and squash the commits and we should be ok to merge that PR and close this one.

imbstack:
Ah yes. I made this PR mostly to test the taskcluster-ci tests. I'll close now.

PR #228, opened 2016-05-03
Add lib-monitor client

gregarndt:


bin/upload-schema.js
```diff
@@ -7,39 +7,39 @@ auto pushing schema's tricky. For now this script does the uploads manually.
...
+var fs = require('fs');
 
-var loadConfig = require('taskcluster-base/config');
-var config = loadConfig({
-  defaults: require('../config/defaults'),
-  profile: require('../config/production'),
-  filename: 'docker-worker'
+var base = require('taskcluster-base');
+var config = base.config({
+  files: [`${__dirname}/../config.yml`],
```
imbstack: Is this line actually necessary? I wouldn't be surprised if you said yes because of some sort of weird thing with the worker, but I haven't seen us needing to use it before.

gregarndt: it appears that the config loading loads config.yml from the current working directory.  So if I'm invoking this outside of the worker directory, then config.yml won't be found.

gregarndt: this is also why we do this in worker.js too.

bin/worker.js
```diff
@@ -176,41 +172,35 @@ async function main () {
...
-        instanceType: config.workerNodeType,
-        provisionerId: config.provisionerId
-      }
-    })
+    mock: profile === 'test'
   });
 
-  config.validator = await base.validator();
-  config.validator.register(require('../schemas/payload'));
+  config.monitor = monitor.prefix(
```
imbstack: I should probably make `monitor.prefix()` take an array of prefixes and join them internally to allow you to avoid concat-ing everything yourself. Not a bug on this review, just a note to myself.

bin/worker.js
```diff
@@ -176,41 +172,35 @@ async function main () {
...
-  }
+  config.monitor.measure('workerStart', Date.now()-os.uptime());
+  config.monitor.count('workerStart');
+
+  config.queue = new taskcluster.Queue({
+    credentials: config.taskcluster
+  });
+
+  config.validator = await base.validator({
+    folder: path.join(__dirname, '..', 'schemas'),
```
imbstack: `base.validator` _should_ handle this correctly by default I think. If it's not working for you without this it might be considered a bug on `tc-lib-validator`.

bin/worker.js
```diff
@@ -176,41 +172,35 @@ async function main () {
...
+    prefix: config.schema.path
+  });
 
   setInterval(
-    reportHostMetrics.bind(this, config),
+    reportHostMetrics.bind(this, {
```
imbstack: Oops, you caught me missing something from the readme. There's a `reportUsage: bool` option to `tc-lib-monitor` that will report cpu/memory usage on an interval. I expect you'll want to set it to false in this project since you have a custom resource reporter.

I'll add this to the readme of tc-lib-monitor.

bin/worker.js
```diff
@@ -314,6 +304,6 @@ async function main () {
 
 main().catch((err) => {
   // Top level uncaught fatal errors!
-  console.error(err);
+  console.error(err.stack);
```
imbstack: `tc-lib-monitor` does something similar on uncaught exceptions https://github.com/taskcluster/taskcluster-lib-monitor/blob/master/src/monitor.js#L23 but also reports it to Sentry. I think we can just keep that and we won't need this `.catch(...)`. I think this might end up silencing the error in Sentry.

config.yml
```diff
@@ -0,0 +1,247 @@
+defaults:
```
imbstack: Every time I see the old `js` config turned into the new `yml` config I get a happy tear in my eye. 💯

lib/docker/docker_image.js
```diff
@@ -86,13 +86,17 @@ export default class DockerImage {
       pullOptions.authconfig = this.credentials;
     }
 
-    return await this.runtime.stats.timeGen(
-        'dockerImageDownloadTime',
-        this.pullImageStreamTo(this.runtime.docker,
-                               dockerImageName,
-                               this.stream,
-                               pullOptions)
+    let start = Date.now();
```
imbstack: I should add a generic `monitor.timethisfunction()` to `tc-lib-monitor`. https://github.com/taskcluster/taskcluster-lib-monitor#timing-handlers is similar but not quite what you want here I guess.

imbstack:
Wow, that required a lot of work to upgrade! I think this looks good. r+

lib/docker/docker_image.js
gregarndt: yea this is pretty much what I was doing here [1] but decided to just rip it all out for now.  I don't do measurements too much in the worker so I took the easy way out.  I was going to port over this stat wrapper to use the new monitor client but just seemed to be unnecessary until the client itself supports a timed handler of sorts.

[1] https://github.com/taskcluster/docker-worker/blob/master/lib/stats/stat.js#L94

gregarndt:
Ok, I addressed your comments and carrying forward the previous r+.  Gonna merge it in.  Let me know if you had any objections.

PR #230, opened 2016-05-20
Expand scopes before validating task scopes

gregarndt:


djmitche:
r+!

lib/task.js
```diff
@@ -94,7 +94,8 @@ Create a list of cached volumes that will be mounted within the docker container
 @param {object} volumes to mount in the container
  */
 async function buildVolumeBindings(taskVolumeBindings, volumeCache, taskScopes) {
```
djmitche: Note that the call to this function wasn't updated to use `await`, which may have been part of the problem.  That was fixed in 518a01a4.

PR #233, opened 2016-06-29
Remove temp task image directory after import

gregarndt:


walac:
r+

test/integration/allow_ptrace_test.js
```diff
@@ -14,7 +14,7 @@ import Debug from 'debug';
 
 let debug = Debug('docker-worker:test:allow-ptrace-test');
 
-suite('allowPtrace feature', () => {
+suite.skip('allowPtrace feature', () => {
```
djmitche: unrelated?

djmitche:
r+ :)

test/integration/allow_ptrace_test.js
gregarndt: Yea, with newer versions of docker testing ptrace in a docker in docker container no longer works, but it does test in a vm just fine.  I removed it here to hopefully reduce noise in the tests for wander's review, but well, that didn't work out either too well.

PR #234, opened 2016-07-14
Update aus4-admin.mozilla.org DNS hack with new IP address

mozbhearsum:
This needs updating after the transition to CloudOps.


rail:
```
$ host aus4-admin.mozilla.org
aus4-admin.mozilla.org is an alias for balrog-aus4-admin.r53-2.services.mozilla.com.
balrog-aus4-admin.r53-2.services.mozilla.com has address 52.26.16.60
```

LGTM

PR #235, opened 2016-07-14
Retry npm install during deployment process

gregarndt:
There are times where there might be an issue contact npm where a retry
would help.  This is specifically useful in the case where there is a
connection reset.

### DIFF
--- a/deploy/packer/app/scripts/deploy.sh
+++ b/deploy/packer/app/scripts/deploy.sh
@@ -26,7 +26,21 @@ mkdir -p $target
 cd $target
 tar xzf $docker_worker_source -C $target --strip-components=1
 sudo chown -R $USER:$USER /home/ubuntu/docker_worker
-npm install --production
+
+retry_count=0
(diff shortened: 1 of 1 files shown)


petemoore:
@gregarndt should we merge this in now? Thanks!

deploy/packer/app/scripts/deploy.sh
```diff
@@ -26,7 +26,21 @@ mkdir -p $target
 cd $target
 tar xzf $docker_worker_source -C $target --strip-components=1
 sudo chown -R $USER:$USER /home/ubuntu/docker_worker
-npm install --production
+
+retry_count=0
+max_retries=5
```
petemoore: should we bump this a little higher just to be super-safe?

petemoore:
From irc:
```
17:56 <pmoore> ah should i merge https://github.com/taskcluster/docker-worker/pull/235 ?
18:00 <garndt> Feel free to merge it
```

PR #236, opened 2016-07-26
Bug 1289511: log instance type, too

djmitche:


djmitche:
I don't know if that test will pass.. we'll see!!

djmitche:
:green_heart:

PR #237, opened 2016-08-06
Add support for Certificate of Trust

gregarndt:


config.yml
```diff
@@ -242,6 +243,7 @@ test:
   dockerVolume: '/tmp'
 
   dockerWorkerPrivateKey: '/worker/test/docker-worker-priv.pem'
+  signingKeyLocation: '/worker/test/fixtures/gpg_signing_key.key'
```
walac: I wonder why we need to specify this configuration twice in the file.

deploy/packer/app.json
```diff
@@ -4,8 +4,8 @@
   "variables": {
     "npmPackage":          "",
     "templateContents":    "",
-    "hvmSourceAMI":        "ami-0710db67",
-    "pvSourceAMI":         "ami-ee17dc8e",
+    "hvmSourceAMI":        "ami-7c22f71c",
+    "pvSourceAMI":         "ami-e422f784",
```
walac: Does this incorporate the recent security fix?

lib/features/certificate_of_trust.js
```diff
@@ -0,0 +1,113 @@
+/**
+The bulk logger writes the task stream directly to disk then uploads that file
+to s3 after the task has completed running.
+*/
```
walac: I think this comment doesn't belong to this module.

lib/features/certificate_of_trust.js
```diff
@@ -0,0 +1,113 @@
...
+    this.file = new temporary.File();
+
+    // Pipe the task stream to a temp file on disk.
+    this.stream = fs.createWriteStream(this.file.path);
+    task.stream.on('data', (d) => {
+      this.hash.update(d);
+    });
+
+    let gzip = zlib.createGzip();
+    task.stream.pipe(gzip).pipe(this.stream);
```
walac: A comment on what this does would be good.

lib/features/certificate_of_trust.js
```diff
@@ -0,0 +1,113 @@
...
+      await uploadToS3(task.queue, task.status.taskId, task.runId,
+                       logStream, 'public/logs/certified.log', expiration, {
+        'content-type': 'text/plain',
+        'content-length': stat.size,
+        'content-encoding': 'gzip'
+      });
+    } catch (err) {
+      debug(err);
+      // Unlink the temp file.
+      await fs.unlink(this.file.path);
```
walac: You can safely `unlink` the file before upload to S3. The stream keeps a file descriptor open, and the file won't be removed before it is closed.

lib/features/certificate_of_trust.js
```diff
@@ -0,0 +1,113 @@
...
+      workerGroup: task.runtime.workerGroup,
+      workerId: task.runtime.workerId,
+      extra: {
+        publicIpAddress: task.runtime.publicIp,
+        privateIpAddress: task.runtime.privateIp,
+        imageHash: task.imageHash,
+      }
+    };
+
+    ['instanceId', 'instanceType', 'region'].forEach((tag) => {
```
walac: nit: the parenthesis around tag are unnecessary.

petemoore:
@escapewindow feel free to provide feedback here too

escapewindow:
@petemoore Thanks for the heads up!  I verified the output in bug 1284991.  I'm not a js expert, so I'll leave the code review to those who are.

config.yml
gregarndt: they are different locations.  One as the default, one for tests.

deploy/packer/app.json
gregarndt: It should?  I pulled down master and rebuilt the base

gregarndt: yup, this was in the base.json in my cot branch:
    {
      "type": "shell",
      "inline": [
        "sudo bash -c 'echo net.ipv4.tcp_challenge_ack_limit = 999999999 >> /etc/sysctl.conf'"
      ]
    }

lib/features/certificate_of_trust.js
gregarndt: quite right, I've updated it.  Thanks for catching that.

gregarndt:
@walac I think this addressed your comments now.

petemoore:
@gregarndt any idea why http://schemas.taskcluster.net/docker-worker/v1/payload.json didn't get updated when this merged?

lib/features.js
```diff
@@ -16,6 +16,15 @@ const features = {
     module: require('./features/local_live_log')
   },
 
+  generateCertificate: {
+    title: 'Enable generation of a certificate for Chain of Trust',
+    description: 'A certificate should be generated which will include information ' +
+                 'for downstream tasks to build a level of trust for the artifacts' +
+                 'produced by the task and the environment it ran in.',
```
petemoore: missing space between `artifacts` and `produced`

test/integration/certificate_of_trust.js
```diff
@@ -0,0 +1,86 @@
...
+    let key = openpgp.key.readArmored(armoredKey);
+    let opts = {
+      publicKeys: key.keys,
+      message: openpgp.cleartext.readArmored(signedCertificate)
+    };
+    let verified = await openpgp.verify(opts);
+
+    assert(verified.signatures[0].valid, 'Certificate does not appear to be valid');
+
+    // computer the hash of the live_backing.log which should be the same as the
```
petemoore: "compute"

PR #238, opened 2016-08-11
Bug 1294205: Update docker-worker AMIs for cve-2016-5696. r=garndt

walac:


deploy/packer/base.json
```diff
@@ -44,6 +44,12 @@
     {
       "type":           "shell",
       "inline":         ["sudo unattended-upgrade"]
+    },
+    {
+      "type": "shell",
```
gregarndt: I definitely like that packer allows us to inline this here, but could we maybe move this to a script where comments can be added as to why we are doing this in the first place?  Just so that if anyone later on looks at this, they know in older kernels to not remove it.

walac: Already discussed on IRC, but commenting here for posterity. The commit already references the cve and the bug, so a comment is redundant.

PR #241, opened 2016-08-23
Update links in schema to point to updated URLs.

SingingTree:
The schema contains two links to the docs which have been updated. The link
about encrypted environment vars was busted, the other link about extended task
graph was okay, due to redirects. I've updated both to fix the busted one and so
they're both consistent.


gregarndt:
thanks!

PR #242, opened 2016-08-23
Attempt to remove files once they are no longer needed

gregarndt:


lib/docker/artifact_image.js
```diff
@@ -78,21 +78,38 @@ export default class ArtifactImage {
...
+      } catch(e) {
+        debug(`Error removing download dir. ${e.stack}`);
+      }
       throw new Error(`Error loading docker image. ${e.message}`);
     }
 
-    await this.renameAndLoad(this.imageName, originalTarball);
-    this.runtime.monitor.measure('task.taskImage.loadTime', Date.now() - start);
+    try {
+      this.runtime.monitor.measure('task.taskImage.loadTime', Date.now() - start);
```
djmitche: doesn't seem like this should be in the try block

djmitche:
So it looks like this mainly moves the removal of the original tarball earlier, before loading into docker.  The rest is, I think, changes to improve the removal on error?

djmitche:
anyway, I don't see any issues here

gregarndt:
Yea I saw a couple issues here, one rather than waiting until the end to remove all the temp files that were created, we can remove them sooner.  Then I noticed that on some instances where loading an image was failing, it wasn't removing those temp files and keeping them around.  Unfortunately the worker does not actively manage that directory on each GC cycle so the files sit around forever and never purged.

PR #243, opened 2016-08-24
Do not setup cancel listener until listeners are consolidated

gregarndt:


jonasfj:
# `r+`

How about adding an error handler other one too?

gregarndt:
Nice suggestion, is this what you were thinking of?

jonasfj:
seems sane to me... even more r+

PR #244, opened 2016-09-02
Remove dependency from aws-sdk-promise. r=jhford

walac:
Amazon has added the .promise() function to aws-sdk and aws-sdk-promise
isn't required anymore. Furthermore, aws-sdk-promise is going to be
deprecated. Details: http://tinyurl.com/hry2fa3

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,8 +20,7 @@
   "bugs": "https://github.com/taskcluster/docker-worker/issues",
   "homepage": "https://github.com/taskcluster/docker-worker",
   "dependencies": {
-    "aws-sdk": "^2.1.5",
-    "aws-sdk-promise": "0.0.2",
+    "aws-sdk": "^2.5.5",
     "azure-storage": "^0.3.0",
     "babel": "^4.7",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -21,7 +21,6 @@
   "homepage": "https://github.com/taskcluster/docker-worker",
   "dependencies": {
     "aws-sdk": "^2.1.5",
-    "aws-sdk-promise": "0.0.2",
```
gregarndt: might need to bump the aws-sdk libarry and make sure it works.

jhford:
I don't know the docker worker codebase very well, but this change looks like the right thing to do.  Also, looking at the output of `git grep "\.data"`, I don't see anything where we use the now-removed `.data` parameter

r+ from me, with the caveat that I don't know the codebase well.

walac:
@gregarndt want to give a last review before merge?

gregarndt:
as far as I'm aware the aws-sdk is only used directly by that upload schema file, so I think this should be ok.

PR #247, opened 2016-09-07
Only run a full gc sweep when no tasks are running

gregarndt:
Most things can be cleaned up while tasks are still running,
such as containers that have exited, and volumes no longer in use.
However, attempting to remove images that might be running or have
layers used by an image currently being loaded will cause pain in the
long run.

### DIFF
--- a/lib/gc.js
+++ b/lib/gc.js
@@ -200,7 +200,6 @@ GarbageCollector.prototype = {
   },
 
   sweep: async function (full=false) {
-    clearTimeout(this.sweepTimeoutId);
     this.emit('gc:sweep:start');
     this.log('garbage collection started');
     await this.markStaleContainers();
--- a/lib/gc.js
+++ b/lib/gc.js
@@ -41,8 +41,6 @@ function GarbageCollector(config) {
   this.log = config.log;
   this.monitor = config.monitor;
   this.taskListener = config.taskListener;
-  // Garbage collection interval in milliseconds
-  this.interval = config.interval || 60 * 1000;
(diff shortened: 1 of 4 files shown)


gregarndt:
@walac  Care to take a look at this? What we're trying to do here is a garbage cycle prior to a worker claiming tasks instead of on 60 second intervals.  We also want to only remove images when no other tasks are running.  This is to prevent image loading/deleting breaking when things are being done at the same time. I hope I didn't mangle the tests too badly here.

gregarndt:
Looking over the failures here, they pass locally and are the same failures we've seen elsewhere.  I fixed the cancel task test which was broken by a different PR

lib/gc.js
```diff
@@ -202,12 +199,7 @@ GarbageCollector.prototype = {
     }
   },
 
-  scheduleSweep: function (interval) {
-    this.sweepTimeoutId = setTimeout(this.sweep.bind(this), interval);
-  },
-
-  sweep: async function () {
-    clearTimeout(this.sweepTimeoutId);
+  sweep: async function (full=false) {
```
walac: Since removing images only when there are no running tasks is a gc policy, I think makes more sense pass the number of running tasks to `sweep` and let the function decide how it wants to handle it.

gregarndt: In this case I just want things that care about a full sweep to do so or not, without needing to know exactly what a number represents.  It seems clearer that the caller of the sweep() method just say if it should perform a full one or not.

gregarndt:
if you're satisfied with this, I would like to get it merged

PR #249, opened 2016-09-08
Bug 1301401 - updating to chainOfTrust version 1 format

petemoore:
Hey @gregarndt I figured the easiest way to test these changes is to open a PR and let taskcluster run the test suite(s). I haven't tested locally, so it might take a few commits to get it right. I'll ping you when it is ready for final review!

CC: @escapewindow @walac

### DIFF
--- a/lib/features.js
+++ b/lib/features.js
@@ -16,13 +16,14 @@ const features = {
     module: require('./features/local_live_log')
   },
 
-  generateCertificate: {
-    title: 'Enable generation of a certificate for Chain of Trust',
-    description: 'A certificate should be generated which will include information ' +
-                 'for downstream tasks to build a level of trust for the artifacts ' +
-                 'produced by the task and the environment it ran in.',
+  chainOfTrust: {
+    title: 'Enable generation of a openpgp signed Chain of Trust artifact',
+    description: 'An artifact named chainOfTrust.json.asc should be generated '
+                 'which will include information for downstream tasks to build '
+                 'a level of trust for the artifacts produced by the task and '
+                 'the environment it ran in.',
     defaults: false,
-    module: require('./features/certificate_of_trust')
+    module: require('./features/chain_of_trust')
   },
 
   // the structure is [name] = { defaults: true/false, module: Handler }
--- a/lib/features.js
+++ b/lib/features.js
@@ -18,9 +18,9 @@ const features = {
 
   chainOfTrust: {
     title: 'Enable generation of a openpgp signed Chain of Trust artifact',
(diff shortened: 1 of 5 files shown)


lib/features.js
```diff
@@ -16,13 +16,14 @@ const features = {
     module: require('./features/local_live_log')
   },
 
-  generateCertificate: {
-    title: 'Enable generation of a certificate for Chain of Trust',
-    description: 'A certificate should be generated which will include information ' +
-                 'for downstream tasks to build a level of trust for the artifacts ' +
-                 'produced by the task and the environment it ran in.',
+  chainOfTrust: {
+    title: 'Enable generation of a openpgp signed Chain of Trust artifact',
```
gregarndt: This is missing the concatenation and will break trying to run the worker.

lib/upload_to_s3.js
```diff
@@ -120,5 +120,5 @@ export default async function uploadToS3 (
     tmp.unlink();
   }
 
-  return `sha256:${digest}`;
+  return `${digest}`;
```
gregarndt: just return digest then, no need to use a template string

lib/features.js
petemoore: Nice spot! Thanks.

lib/upload_to_s3.js
petemoore: Ah yes! Thanks.

petemoore:
@gregarndt Tests have finished, we have 38 failures - are they genuine problems related to this PR? Thanks!

https://public-artifacts.taskcluster.net/V6iUaCaUQjOpRThfyQnnHQ/0/public/logs/live_backing.log

gregarndt:
I closed/opened the PR to retrigger another graph creation.  Retriggering an individual task doesn't work because right now we're using encrypted environment variables that encode the task ID into them.

petemoore:
@gregarndt can you check https://tools.taskcluster.net/task-inspector/#mMF4b_SSTFCHzyz6jViRlg/0 and https://tools.taskcluster.net/task-inspector/#Tpt4eVdLTvWL3b1nqXg1IQ/0 to see if you think these are related. I think these might also be timeout/socket issues.

gregarndt:
So it looks like there are 2 unique errors, one of them wasn't caused by what you did from what I can see, and the other one I fixed in another PR.  I think you're good here.

petemoore:
@gregarndt Great! Does that mean we can merge?

gregarndt:
Ok, merged, but won't be deployed immediately.  It's Friday so I would prefer to hold off to land changes on Monday.

PR #250, opened 2016-09-09
Refactor to use queue.claimWork

jonasfj:
Okay, this is more of a "I have no clue what I'm doing -- have look maybe this works" kind of hack...

I would love for us to try and use `queue.claimWork` in production with docker-worker. Ideally, we can try it for a high volume workerType before we roll it out to everything.

But I would really like to get some load on this, so we can show that it works and that our heroku nodes won't fallover from having to do so much polling work... Hopefully it works, if it doesn't we might have to roll back, so we should have old AMIs handy when we try this out.

I suspect tests will fail as the scope `queue:claim-work:<provisionerId>/<workerType>` is required for `queue.claimWork`.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -123,12 +123,6 @@ defaults:
     # Ideally we want to keep the period less than double the average task but more than the runtime of
     # most (say, ~75%) of tasks.
     slowdownDivisor: 3
-    # If signed url for queue expires within now()+expiration, refresh queues
-    expiration: 300000
-    # Number of times to retry requests to the task queue
-    maxRetries: 5
-    # Amount of time to wait between retries
-    requestRetryInterval: 2000
 
   #Registries which we can authenticate against for pulls:
   #  registries: {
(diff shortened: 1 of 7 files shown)


walac:
World is down, all tests failed :)

jonasfj:
any idea what went wrong? I'm clueless here...

gregarndt:
There's quite a few different issues here that I didn't see on the last PR merged, but I don't think it's really related to this work.  Retriggering a task graph to take a look

walac:
I'll take a look why all tests fail

gregarndt:
I'm going to check out the tests after they run, I ran some locally and they were fine.  npm shrinkwrap wasn't updated to include the new taskcluster client, and the role that that's used for tests did not have the new claim-worker or worker-id scopes that are needed now.  updated those too.

gregarndt:
Ok, I've fixed this up enough to where it should be ok.

PR #251, opened 2016-09-09
Support lz4 compressed images

gregarndt:


lib/docker/artifact_image.js
```diff
@@ -15,6 +16,21 @@ import sleep from '../util/sleep';
 
 let debug = Debug('docker-worker:artifactImage');
 
+async function decompressLz4File(inputFile) {
+  let outputFile = path.join(path.dirname(inputFile), path.basename(inputFile, '.lz4'));
```
walac: Isn't this the same as `inputFile + '.lz4'`?

lib/docker/artifact_image.js
```diff
@@ -15,6 +16,21 @@ import sleep from '../util/sleep';
...
+  // decompress using lz4
+  let input = fs.createReadStream(inputFile);
+  let output = fs.createWriteStream(outputFile);
+  let decoder = lz4.createDecoderStream();
+  input.pipe(decoder).pipe(output);
+
+  await new Promise((accept, reject) => {
+    decoder.on('end', accept);
+    decoder.on('error', reject);
+  });
```
walac: This theoretically creates a potential deadlock. Imagine after the call to `pipe` it immediately raises, say, the `error` event, before `decoder.on('error', reject)` is called, then the promise would never be rejected. Although in practice this is really hard to happen, if not impossible, I suggest changing it to this:

```javascript
let promise = new Promise((accept, reject) => {
  decoder.on('end', accept);
  decoder.on('error', reject);
});
input.pipe(decoder).pipe(output);
await promise;
```

lib/docker/artifact_image.js
```diff
@@ -62,24 +78,38 @@ export default class ArtifactImage {
...
           this.runtime.dockerConfig
       );
       this.runtime.monitor.measure('task.taskImage.downloadTime', Date.now() - start);
 
-      await this.renameAndLoad(this.imageName, originalTarball);
+      switch (path.extname(downloadedFile)) {
+        case '.lz4':
+          this.stream.write(fmtLog('Decompressing downloaded image'));
+          tarballPath = await decompressLz4File(downloadedFile);
+          fs.unlink(downloadedFile);
```
walac: In case of `decompressLz4File` fails, `downloadFile` but get stuck in the fs. I suggest using `try`/`finally` to guarantee the file is unliked.

lib/docker/artifact_image.js
gregarndt: Ah darn it, you're right, this should be changed around.  Good catch.

lib/docker/artifact_image.js
gregarndt: basename strips the provided extension off if ti's the last extension of the filename.  So path.basename('image.tar.lz4', '.lz4') becomes 'image.tar'

lib/docker/artifact_image.js
gregarndt: if decompress fails, an exception should be thrown that should be caught by the outter try block which removes the entire directory

lib/docker/artifact_image.js
walac: Alternatively you could also do:

```javascript
  decoder.on('end', accept);
  decoder.on('error', reject);
  input.pipe(decoder).pipe(output);
});
```

lib/docker/artifact_image.js
walac: Oh, yeah.

lib/docker/artifact_image.js
walac: Nice then.

gregarndt:
@walac I changed to using child_process to decompressing so we can have the speed of precompiled C libraries rather than node stuff.  Please take another look!

lib/docker/artifact_image.js
```diff
@@ -15,6 +16,46 @@ import sleep from '../util/sleep';
...
+    proc.on('exit', accept);
+  });
+
+  if (proc.exitCode !== 0) {
+    throw new Error(
+        'Could not decompress image file.' +
+        `Exit Code: ${proc.exitCode} Errors: ${err.join('\n')}`);
+  }
+  return outputFile;
+}
```
walac: lines 23 to 36 could be isolated in a helper function and called by `decompressLz4File` and `decompressZstdFile`

walac:
Several tests failing, not sure if it is related to this pr or not.

gregarndt:
Yea I checked them out and didn't see anything standing out as caused by this PR.

lib/docker/artifact_image.js
gregarndt: Yea I definitely got that same feeling that we could do that, but given it's copied twice, I was holding off if we did it again.

PR #252, opened 2016-09-16
Efficiency monitoring

jonasfj:
@gregarndt, have a look see if this is crazy... I'm intentionally only recording per `<provisionerId>.<workerType>` so that we have something that's easy to compare.

I suggest you just try to roll this our early in the morning, and kill all workers so we have some numbers..
The desktop-test workertypes are probably the biggest load.

### DIFF
--- a/lib/task_listener.js
+++ b/lib/task_listener.js
@@ -29,7 +29,16 @@ export default class TaskListener extends EventEmitter {
     this.lastTaskEvent = Date.now();
     this.host = runtime.hostManager;
     this.supersedingTimeout = 5000;
-
+    this.lastKnownCapacity = 0;
+    this.lastCapacityState = {
+      time: new Date(),
+      idle: this.lastKnownCapacity,
+      busy: this.runningTasks.length,
+    };
+    this.reportCapacityStateIntervalId = setInterval(
+      this.reportCapacityState.bind(this), 60 * 1000,
+    );
+    this.capacityMonitor = this.runtime.workerTypeMonitor.prefix('capacity');
     this.deviceManager = new DeviceManager(runtime);
   }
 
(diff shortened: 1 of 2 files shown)


lib/task_listener.js
```diff
@@ -328,6 +335,59 @@ export default class TaskListener extends EventEmitter {
...
+    if (this.lastCapacityState.busy >= 3) {
+      this.capacityMonitor.count('running-ge-3', time);
+    }
+    if (this.lastCapacityState.busy >= 4) {
+      this.capacityMonitor.count('running-ge-4', time);
+    }
+    if (this.lastCapacityState.busy >= 6) {
+      this.capacityMonitor.count('running-ge-6', time);
+    }
+    if (this.lastCapacityState.busy >= 8) {
```
gregarndt: Why the arbitrary limit of 8 here? Also, do we not care about when it's >= 5 or >= 7? :) Same goes for below this.

lib/task_listener.js
```diff
@@ -328,6 +335,59 @@ export default class TaskListener extends EventEmitter {
     if (this.isIdle()) this.emit('idle', this);
   }
 
+  reportCapacityState() {
+    let state = {
+      time: new Date(),
+      idle: this.lastKnownCapacity,
+      busy: this.runningTasks.length,
+    };
+    let time = (state.time.getTime() - this.lastCapacityState.getTime()) / 1000;
```
jonasfj: nice catch,

lib/task_listener.js
jonasfj: This is just to get some insight.. mostly we have capacity 4...

jonasfj:
@gregarndt, okay, this should help.

PR #257, opened 2016-10-24
Env vars should always be strings

petemoore:
This avoids unexpected translations.

Example:

```
"Height": 1.95
```

could result in the env variable `Height` having value `1.95` or `1,95` depending on locale. Similarly, `1000000000` might be represented as `1,000,000,000` or `1E9` or `1.000E9` as a string, depending on how the worker translates numbers to strings. Since env vars are necessarily strings, it is not easy to know how they will get translated.

Lastly, how would `"ENV_X": { "A": "B" }` be interpreted as the environment variable `ENV_X`? It is not easy to say. Maybe raw json?

Enforcing that env vars are specified as strings avoids this ambiguity, and strange bugs arising as a result of an unexpected conversion, in rare edge cases.

### DIFF
--- a/schemas/payload.json
+++ b/schemas/payload.json
@@ -195,9 +195,12 @@
       }
     },
     "env": {
-      "title": "Environment variable mappings.",
+      "title": "Environment variable mappings. Must be strings.",
       "description": "Example: ```\n{\n  \"PATH\": '/borked/path' \n  \"ENV_NAME\": \"VALUE\" \n}\n```",
-      "type": "object"
+      "type": "object",
+      "additionalProperties": {
+        "type": "string"
+      }
     },
     "maxRunTime": {
       "type": "number",


petemoore:
@gregarndt There seem to be a lot of test failures, but I'm not sure they relate to this change - what is your opinion?

schemas/payload.json
```diff
@@ -195,9 +195,12 @@
       }
     },
     "env": {
-      "title": "Environment variable mappings.",
+      "title": "Environment variable mappings. Must be strings.",
       "description": "Example: ```\n{\n  \"PATH\": '/borked/path' \n  \"ENV_NAME\": \"VALUE\" \n}\n```",
-      "type": "object"
+      "type": "object",
+      "additionalProperties": {
+        "type": "string"
```
gregarndt: Are we sure that this will not break elsewhere? I'm not sure if gecko tasks specify a mixture of strings and numbers.

gregarndt:
I don't think these failures were caused by your changes.

schemas/payload.json
petemoore: Good point. I think we can generate a full gecko task graph, and then inspect it. We could also do the same for NSS.

@jonasfj how much work is it to dig through existing task definitions in Azure and see if any of them have non-strings for env vars? I'm wondering if this is easier than studying all the task graph generators...

jonasfj: It's non-trivial...  there is some logic here:
https://github.com/taskcluster/taskcluster-queue/blob/master/src/scan.js

But be very care not to accidentally delete everything!

IMO, it's safer and easier to just record all tasks using pulse exchanges... and then compare with a days worth of tasks... Or list a few task groups from treeherder...

`queue.listTaskGroup` can list everything in a group/push.

gregarndt:
Since this could be parked for some time, I'm going to propose closing the PR and we can reopen it once we know it can land.

PR #259, opened 2016-11-06
Fix issue with expanding task scopes

gregarndt:
* Restricted tests to docker-worker:cache:garbage-test* instead of the global docker-worker:cache:*
* Used new taskcluster-lib-scopes package instead of using deprecated utilities from taskcluster-base


gregarndt:
@jonasfj Mind double checking how I'm doing the scopeMatch here and the test I added for it?  Basically we changed from using taskcluster-base to taskcluster-lib-scopes, and also added the ability to expand the scopes before checking for times where a task might use a role in task.scopes.

lib/util/scopes.js
```diff
@@ -11,7 +12,10 @@ import { scopeMatch } from 'taskcluster-base/utils';
...
  */
-export function hasPrefixedScopes(prefix, resources, taskScopes) {
-    let neededScopes = Object.keys(resources).map((r) => { return prefix + r; });
-    return scopeMatch(taskScopes, [neededScopes]);
+export async function hasPrefixedScopes(prefix, resources, taskScopes=[]) {
+  let auth = new taskcluster.Auth();
+  let expandedScopes = await auth.expandScopes({scopes:taskScopes});
+  let neededScopes = Object.keys(resources).map((r) => { return prefix + r; });
+
+  return scopeUtils.scopeMatch(expandedScopes.scopes, [neededScopes]);
```
djmitche: :+1: on this bit from me :)

gregarndt:
Ok, thumbs up by dustin for the actual code changes that would impact things, going to merge based on that.  The other changes were just to the tests.

PR #260, opened 2016-11-08
Improved error message

jonasfj:
Found this when I used `.tar.zstd` instead of `.tar.zst` my bad...

No need to rebuild AMIs, I'm happy if this comes out whenever we roll the next update..


gregarndt:
Good call, thank you!

PR #261, opened 2016-11-14
Not all systems have bash, busybox only has sh

jonasfj:


gregarndt:
Fine for me.  How urgently do you need this deployed?

jonasfj:
@gregarndt, no need to deploy anytime soon, we have bash on all images in practice.

Note: we should check that interactive works after we've rolled this change.

gregarndt:
tested this out using one of our mochitests on the ami-test worker type, seemed ok.

PR #272, opened 2016-12-20
Purge timestamp should be JSON

gregarndt:


gregarndt:
@imbstack had to fix up a coupel of problems with the way I originally implemented the purge cache stuff.  Mind taking a look?

gregarndt:
There was just one test removed, and it was testing a situation that is no longer possible due to the change in when we check for purging of the caches.  We only purge caches during a garbage collection event now, and that only happens prior to claiming tasks.  Previously we would be responding to purge pulse events while executing tasks, this was the test for that deprecated scenario.

PR #273, opened 2016-12-22
Bug 1299457: include runId in apparmor profile name

djmitche:
This avoids conflicts if a task is re-run on the same instance.


djmitche:
(the re-triggered run failed with deadline-exceeded)

djmitche:
OK, I think those failures are benign?

gregarndt:
The failures mostly appear to be related to some volume cache tests, which were unintentionally broken by a few commits back and should have been fixed up by ba4b110312f93258d7c29d36a64b3b217a771fa7

gregarndt:
Note: this change has been merged, but not yet deployed.  Since the next AMI deployment will contain a few different bugs fixed, we should test it out on a non-production worker type before widespread deployment.

PR #275, opened 2017-01-10
Use livelog v4

jonasfj:
This allows cors requests to see transfer-encoding... so logviewer can detect that the log is live.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -54,7 +54,7 @@ defaults:
 
   # Image used to  create the taskcluster proxy container.
   taskclusterProxyImage: 'taskcluster/taskcluster-proxy:latest'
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -54,7 +54,7 @@ defaults:
 
   # Image used to  create the taskcluster proxy container.
   taskclusterProxyImage: 'taskcluster/taskcluster-proxy:latest'
-  taskclusterLogImage: 'taskcluster/livelog:v3'
+  taskclusterLogImage: 'taskcluster/livelog:v4'
```
petemoore: Note, we should probably align on the version number in the software, and the docker tag - e.g. this should probably be v1.0.1 rather than v4.

PR #276, opened 2017-01-10
Pull some of the basic sidecar images

gregarndt:


deploy/packer/app/scripts/deploy.sh
```diff
@@ -1,5 +1,11 @@
 #! /bin/bash -vex
 
+# Pull images used for sidecar containers
+docker pull taskcluster/taskcluster-proxy:latest
+docker pull taskcluster/livelog:v3
```
jonasfj: I just landed a PR upgrading to v4 today... hehe :)

gregarndt:
I need to revert this, it does not do what I intended.  It does pull down the images during AMI creation, but it's pulled to an ephemeral drive.  When a worker comes up, that drive is empty and it will download it again.

PR #279, opened 2017-01-19
Bug 1307509 - Load sidecar images during startup

gregarndt:


deploy/template/etc/init/docker-worker.conf
```diff
@@ -18,11 +18,14 @@ script
     . /etc/default/$UPSTART_JOB
   fi
 
-  # Pull images used for sidecar containers
-  docker pull taskcluster/taskcluster-proxy:latest
-  docker pull taskcluster/livelog:v4
-  docker pull taskcluster/dind-service:v4.0
-  docker pull taskcluster/relengapi-proxy:2.0.1
+  echo "Loading docker sidecar images"
```
djmitche: I'm confused how this will change things for the user: if we were already docker-pulling these at startup, why does it help to docker load them instead?

gregarndt: The shift here is rather than pulling the the sidecar images during the first task that is claimed, we are doing so at worker startup.  The tarball for the images is saved in the AMI, and then loaded prior to docker-worker claiming any tasks.  Ideally this means that we are reducing the necessity of the docker registry for a task if the sidecar images are already loaded when a task is claimed, and an image for a task is pulled from a artifact.

djmitche: I see, so the `docker pull` commands removed in this hunk sometimes failed, but `docker-worker.conf` continued executing and started the worker, which then tried to pull again, and failed again.

PR #282, opened 2017-01-20
Bug 1332490 - Upgrade docker 1.12

gregarndt:
This will get us on the latest stable version of docker (1.13.0 is out now but I don't consider their .0 releases stable.

This will address some security issues as well as a race condition in docker that we might have been seeing.


test/integration/file_artifact_test.js
```diff
@@ -222,7 +222,7 @@ suite('artifact extration tests', () => {
       }
     });
 
-    let errorMessage = '[taskcluster] Artifact "public/my-missing.txt" not found at "/this-file-is-missing.txt"';
```
gregarndt: timestamps are included in the prepended text and this test isn't specifically testing that it's added so might as well not check.

gregarndt:
Yup, that pretty much sums it up.  The tests we do have for docker-worker came back all green so I think that's a good step towards testing this out on a worker type.  This change will roll out separately from the latest merges to docker worker just so we can monitor what's going on.

PR #283, opened 2017-01-25
Add claimWork workflow

gregarndt:


gregarndt:
this should make @jonasfj happier

PR #284, opened 2017-01-31
Remove Availability Zone letter from workerGroup

gregarndt:


gregarndt:
@jonasfj Having the AZ letter when recording the metrics introduces a lot of unique series names.  I don't think we ever care about this.

PR #286, opened 2017-02-27
Bug 1341743: record artifact upload metrics. r=garndt,jhford

walac:
Record time, number of bytes and transfer rate for each artifact
uploaded.

@jhford @gregarndt How can I test it?

### DIFF
--- a/src/lib/features/artifacts.js
+++ b/src/lib/features/artifacts.js
@@ -9,6 +9,7 @@ import mime from 'mime';
 import tarStream from 'tar-stream';
 import Debug from 'debug';
 import Promise from 'promise';
+import assert from 'assert';
 
 import {fmtLog, fmtErrorLog} from '../log';
 import uploadToS3 from '../upload_to_s3';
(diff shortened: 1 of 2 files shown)


jhford:
@gregarndt i can't reply inline, but I dont' really care whether times are in ms or s or bytes/ms or bytes/s, fwiw

gregarndt:
So it looks like this PR has become stale.  Do we still need it?  If so, let's get @jhford to approve the review as well and get it deployed.  (it'll need a rebase since there are some merge conflicts)

src/lib/features/artifacts.js
```diff
@@ -139,8 +139,17 @@ export default class Artifacts {
...
+        let start = Date.now();
+        let {digest, size} = await uploadToS3(taskHandler.queue, taskId, runId, stream,
                          entryName, expiry, headers);
+
+        let monitor = taskHandler.runtime.monitor;
+        let elapsedTime = (Date.now() - start) / 1000; // in seconds
+
+        monitor.measure('artifacts.size', size);
+        monitor.measure('artifacts.uploadTime', elapsedTime);
+        monitor.measure('artifacts.throughput', size / elapsedTime); // Bytes/sec
```
jhford: is it possible to have this as a per-region and also a global metric?  Since we're aggregating the data points *before* sending them to SignalFX, the only chance we have to tell the difference between global and per-region is if we note it before sending to statsum here.

src/lib/upload_to_s3.js
```diff
@@ -46,6 +47,9 @@ export default async function uploadToS3 (
     }
     debug(`wrote source file to ${tmp.path} for ${artifactName}`);
 
+    const stat = fs.statSync(tmp.path);
```
jhford: if we're in an async function already, have we considered importing `mz/fs`?  For this case, it's probably not worth the concern, but it might be a good thing to have since we already have the overhead of an async function.

src/lib/features/artifacts.js
```diff
@@ -139,8 +140,29 @@ export default class Artifacts {
...
+
+        let monitor = taskHandler.runtime.monitor;
+        let start = process.hrtime();
+
+        let {digest, size} = await uploadToS3(taskHandler.queue, taskId, runId, stream,
                          entryName, expiry, headers);
+
+        // save the time taken to upload the artifact
+        let elapsedTime = process.hrtime(start);
+        let uploadTime = elapsedTime[0] + elapsedTime[1] / 1e9; // in seconds
```
jhford: Any reason not to measure in milliseconds as we do in lib-monitor for other datapoints?

walac: Nope, I just did in secs because it was specified in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1341743#c0).

jhford: let's keep speed in bytes/second and report the time in milliseconds.  With that, r+

walac:
There is one failing test that I ran locally and passed. I confirm it is not related to this patch.

PR #287, opened 2017-03-24
Bug 1346251 Prefer newer builtin methods over lodash

stanleyye:
Couldn't find too many builtin methods > lodash. There were a some lodash methods such as defaults and deepDefaults but its a bit different than something like Object.create().

### DIFF
--- a/src/lib/features/artifacts.js
+++ b/src/lib/features/artifacts.js
@@ -4,7 +4,6 @@ deals with the extract of both single and multiple artifacts from the docker
 container.
 */
 
-import _ from 'lodash';
 import mime from 'mime';
 import tarStream from 'tar-stream';
 import Debug from 'debug';
@@ -181,14 +180,14 @@ export default class Artifacts {
     if (typeof artifacts !== 'object') return;
 
     // Upload all the artifacts in parallel.
-    await Promise.all(_.map(artifacts, (value, key) => {
+    await Promise.all(artifacts.map((value, key) => {
       return this.uploadArtifact(taskHandler, key, value).catch((err) => {
         errors[key] = err;
       });
     }));
 
     if (Object.keys(errors).length) {
-      _.map(errors, (value, key) => {
+      errors.map((value, key) => {
         debug('Artifact upload %s failed, %s, as JSON: %j', key, value, value, value.stack);
       });
 
--- a/src/lib/features/artifacts.js
+++ b/src/lib/features/artifacts.js
@@ -4,6 +4,7 @@ deals with the extract of both single and multiple artifacts from the docker
 container.
 */
 
+import _ from 'lodash';
 import mime from 'mime';
 import tarStream from 'tar-stream';
 import Debug from 'debug';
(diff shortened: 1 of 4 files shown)


test/integration/encrypted_env_variables_test.js
```diff
@@ -37,7 +37,7 @@ suite('encrypted private env variables', () => {
 
   function getEncryptedEnvPayload(payloadData) {
 
-    return Promise.all(_.map(payloadData, (data) => {
+    return Promise.all(payloadData.map((data) => {
```
gregarndt: Could you have a look at the test files where you removed lodash and see if the import of lodash could be removed?

stanleyye: sure, it seems like I missed that

src/lib/private_key.js
```diff
@@ -95,7 +94,7 @@ PrivateKey.prototype = {
 
     // For each encrypted variable, create a promise and wait for all
     // promises to complete
-    return Promise.all(_.map(taskPayload.encryptedEnv, function(encryptedVar) {
+    return Promise.all(taskPayload.encryptedEnv.map(function(encryptedVar) {
```
gregarndt: This will thrown an error if taskPayload.encryptedEnv is undefined

src/lib/features/artifacts.js
```diff
@@ -181,14 +180,14 @@ export default class Artifacts {
     if (typeof artifacts !== 'object') return;
 
     // Upload all the artifacts in parallel.
-    await Promise.all(_.map(artifacts, (value, key) => {
+    await Promise.all(artifacts.map((value, key) => {
```
gregarndt: artifacts is an object, which does not have a map() method.

src/lib/features/artifacts.js
```diff
@@ -181,14 +180,14 @@ export default class Artifacts {
...
       return this.uploadArtifact(taskHandler, key, value).catch((err) => {
         errors[key] = err;
       });
     }));
 
     if (Object.keys(errors).length) {
-      _.map(errors, (value, key) => {
+      errors.map((value, key) => {
```
gregarndt: errors is an object, which does not have a map() method

stanleyye:
Let me get back to you on this later, I was having trouble running the tests.

gregarndt:
Running tests can be problematic locally if you do not have TaskCluster credentials.  It might be easiest to make changes and push them up to the PR to let CI capture them.

stanleyye:
sweet, that would be ideal. Is there any way for CI to automatically capture the changes? It seems like you have to close and re-open the PR? As well how do you know what errors there are from the graphs? I can't find the page where the test errors are displayed.

gregarndt:
It's not all the time where a close/open triggers CI, but sometimes it helps to kick start it if it didn't automatically start for some reason.  Let's see what happens the next time you push.  If things work well, you should see a comment like: https://github.com/taskcluster/docker-worker/pull/287#issuecomment-294929948

that link will take you to a TaskCluster group inspector, which can a little confusing to get to the results you need (improvements to it are in the pipeline).  From there you will have a list of tasks, you can select any one of them, then click the menu for "Runs", select the latest run, and click the Logs link.

src/lib/features/artifacts.js
stanleyye: is it best to just leave this part alone? I was thinking of Object.keys(artifacts)..map(function(key, index) { .... }) but then you have to call the accessor like artifacts[key] for the value. With lodash, its much more simpler. What do you think?

stanleyye:
hey @gregarndt, it seems the tests fail with no changes made?

gregarndt:
hey @stanleyye , sorry that you're seeing failures.  Unfortunately some of these are flakey and result in some false errors.  If you submit your recent changes I'm happy to sort through the errors and run them locally to see what is truly a failure.  This isn't ideal, I admit, but it'll work until time is invested in making the tests less error prone.  Thanks for your persistence and patience!

stanleyye:
hi @gregarndt, could you try testing my latest changes?

stanleyye:
it looks like everything passed? not sure if some other small changes i made were worthwhile.

gregarndt:
Sorry for the long delay on this, but you're right, it looks like things are ok! Thanks for your patience on this PR and for your help getting it landed!

PR #292, opened 2017-05-26
Bug 1297151: Proposal to kill with sigint first

jonasfj:
This is just a quick proposal... I don't really want to be responsible for this, as I don't have tests running locally.

But this could be all that is needed.


gregarndt:
I don't

jonasfj:
bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1297151

src/lib/task.js
```diff
@@ -696,11 +696,20 @@ export class Task extends EventEmitter {
...
+      // But first we kill with SIGINT, giving the process a chance to cleanup
+      setTimeout(() => {
+        this.dockerProcess.kill();
+        this.stream.write(fmtErrorLog(
+          'Task timeout after %d seconds. Force killing container.',
+          this.task.payload.maxRunTime
+        ));
+      }, 5 * 1000);
+      try {
+        this.dockerProcess.container.kill({signal: 'SIGINT'}).catch(err =>
```
gregarndt: Will this propagate to child processes?  I think that if bash is pid 1 for the container, things aren't going to get the SIGINT signal.

jonasfj: I think propagation and graceful termination is the responsibility of the task-author...
Notably: we should **not** be sending `SIGINT` to all subprocesses, because it is the job of the task-author to gracefully stop...

If the task script doesn't gracefully stop, which could involve stopping some processes before others, etc... then we will force kill..

If we send SIGINT to every subprocess the root process won't be able to make a controlled graceful termination, because we might have killed resources it depended on for that purposes.

gregarndt: ok, so it's understood that this is only the start to the fix.  The other part will be the task to capture the signal and try to gracefully do things, which is probably the largest bit to do overall.

jonasfj: yes, and it's really only for tasks that cares about this stuff...

It might be something gps will play with in `run-task` to ensure that mercurial cases aren't locked or anything bad...

PR #293, opened 2017-06-29
Added hooks for idle work...

jonasfj:
Do we agree that `startIdleWork` and `stopIdleWork` is called in the right places?

Worker will block the next task until `stopIdleWork` is done... and `startIdleWork` will be called whenever we're idle, so if idle work isn't running it can start it... and it can measure that it's been idle for 2 min before starting idle work too... etc...

Agree, that this is what we need to fit the logic into? Or are we missing some hooks, are we not blocking correctly before running new tasks? or something else I'm missing... I think I managed to ensure `stopIdleWork` gets called when we shutdown...

### DIFF
--- a/src/lib/task_listener.js
+++ b/src/lib/task_listener.js
@@ -10,6 +10,7 @@ import request from 'superagent-promise';
 import { Task } from './task';
 import { EventEmitter } from 'events';
 import { exceedsDiskspaceThreshold } from './util/capacity';
+import IdleWork from './idlework';
 
 const debug = Debug('docker-worker:task-listener');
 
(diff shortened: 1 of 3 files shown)


src/lib/task_listener.js
```diff
@@ -331,7 +334,10 @@ export default class TaskListener extends EventEmitter {
     this.runningTasks.splice(taskIndex, 1);
     this.lastKnownCapacity += 1;
 
-    if (this.isIdle()) this.emit('idle', this);
+    if (this.isIdle()) {
```
gregarndt: As you described we probably want to only call start idle work unless it's been idle for a certain amount of time since we call claimWork very frequently.

gregarndt:
is this in a state that should be reviewed again?

jonasfj:
No I think we need to work out why it's broken first...

Den 28. jul. 2017 3.13 PM skrev "Greg Arndt" <notifications@github.com>:

PR #297, opened 2017-07-28
Move to taskcluster-github for CI

gregarndt:
This PR moves from using a custom heroku app for triggering CI on pushes to using taskcluster-github.  The main change is how test creds get into the test environment to use for the integration tests.  Previously this was done using encrypted env variables that the heroku app injected, but now we'll move to using github secrets.


gregarndt:
Looks like the only failure that happened was caused by retrying directory uploads, which I don't think is related to this PR...changes are it was intermittently failing before.

.taskcluster.yml
```diff
@@ -46,4 +46,198 @@ tasks:
...
+    extra:
+      github:
+        env: true
+        events:
+          - push
+          - pull_request.*
+
+  - provisionerId: aws-provisioner-v1
+    workerType: worker-ci-test
+    retries: 0
```
walac: `retries` is 0 but the failed task ran 3 times, did you retrigger manually?

walac:
In time, I filed a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1385920) to investigate the failed test.

gregarndt:
As part of this review, can you both check out the scopes assigned to the client in the ci-creds secret?

gregarndt:
Some updates have been made to the ci creds to remove unnecessary scopes and things are still passing so we'll consider this good to go.  An attempt was made to try to use the proxy instead of injecting creds, but because of some operations within the worker relying on building a signed URL, that does not work since credentials are needed

PR #299, opened 2017-08-02
Do not use primary task credentials when resolving superseded task

gregarndt:
This should fix resolving superseded tasks.  Currently without the fix, a scope error will be thrown because the queue client being used is for the primary task which does not have scopes to resolve the superseded task.

### DIFF
--- a/src/lib/task.js
+++ b/src/lib/task.js
@@ -221,7 +221,7 @@ export class Reclaimer {
     this.log('reclaiming task');
 
     try {
-      let queue = this.task.createQueue(this.claim.credentials, this.runtime);
+      let queue = this.task.createQueue(this.claim.credentials);
       this.claim = await queue.reclaimTask(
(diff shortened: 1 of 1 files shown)


src/lib/task.js
```diff
@@ -1012,7 +1012,7 @@ export class Task extends EventEmitter {
 
   @return New queue.
   */
-  createQueue(credentials, runtime) {
+  createQueue(credentials) {
```
djmitche: why does this need runtime now?

gregarndt: It doesn't need it now, I removed runtime from being passed in because it was not needed.  Maybe at some point in the past it was needed but that has since changed.

Side note: I looked into why we need/want this method at all, and it seems it's used for injecting a queue client in tests by overriding this method so probably simplest just keeping it here.

djmitche: Uhh, how did I read that backward..

PR #300, opened 2017-08-03
Fix json schema to use "format": "uri" rather than "url"

petemoore:
Just a minor papercut. See [logs](https://papertrailapp.com/groups/853883/events?q=unknown%20format%20%22url%22%20ignored%20in%20schema%20at%20path) and [json schema validation specification](http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.6).


petemoore:
@djmitche (not sure why schema validation on publication didn't pick this up) - any ideas?

petemoore:
@jonasfj @jhford do either of you know how the schema validation is performed, and if we get require [defined](http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3) formats?

PR #301, opened 2017-08-03
Fixed the spec to match the implementation

petemoore:
I confirmed with @dividehex that the RelEng superseding service also considers tasks to be sorted from most recent to least recent, and thus it was a spec error, rather than an implementation error.


petemoore:
@dividehex fyi

djmitche:
:heart: doc updates especially when they're correcting bogus info

PR #304, opened 2017-08-16
Webhooktunnel Support [WIP][No merge]

ckousik:
Initial webhooktunnel support. Work in progress.

### DIFF
--- a/src/bin/worker.js
+++ b/src/bin/worker.js
@@ -16,6 +16,7 @@ import GarbageCollector from '../lib/gc';
 import VolumeCache from '../lib/volume_cache';
 import PrivateKey from '../lib/private_key';
 import ImageManager from '../lib/docker/image_manager';
+import WebhookServer from '../lib/webhookserver';
 
 // Available target configurations.
 var allowedHosts = ['aws', 'test'];
@@ -238,6 +239,17 @@ program.parse(process.argv);
   shutdownManager.observe(taskListener);
 
   await taskListener.connect();
+
+  // Set up the webhookServer instance for runtime.
+  var webhookServer = null;
+  try{
+    webhookServer = new WebhookServer(config.taskcluster);
+    let url = await webhookServer.start();
+    debug("webhookserver connected. hosted on "+url);
+  }catch(e){
+    debug("could not set up webhookserver: error: "+ e);
+  }
+  runtime.webhookServer = webhookServer;
 
   runtime.log('start');
 
--- a/src/bin/worker.js
+++ b/src/bin/worker.js
@@ -241,15 +241,7 @@ program.parse(process.argv);
   await taskListener.connect();
 
   // Set up the webhookServer instance for runtime.
-  var webhookServer = null;
-  try{
-    webhookServer = new WebhookServer(config.taskcluster);
-    let url = await webhookServer.start();
-    debug("webhookserver connected. hosted on "+url);
-  }catch(e){
-    debug("could not set up webhookserver: error: "+ e);
-  }
-  runtime.webhookServer = webhookServer;
+  runtime.webhookServer = await WebhookServer.startServer(config.taskcluster);
 
   runtime.log('start');
 
--- a/src/bin/worker.js
+++ b/src/bin/worker.js
@@ -233,15 +233,15 @@ program.parse(process.argv);
     verifySSLCertificates(runtime);
   }
 
+  // Set up the webhookServer instance for runtime.
+  runtime.webhookServer = await WebhookServer.startServer(config.taskcluster);
+
   // Build the listener and connect to the queue.
   var taskListener = new TaskListener(runtime);
   runtime.gc.taskListener = taskListener;
   shutdownManager.observe(taskListener);
 
   await taskListener.connect();
-
-  // Set up the webhookServer instance for runtime.
-  runtime.webhookServer = await WebhookServer.startServer(config.taskcluster);
 
   runtime.log('start');
 
--- a/src/lib/features/webhooklogs.js
+++ b/src/lib/features/webhooklogs.js
@@ -98,7 +98,7 @@ class WebhookLogs {
     this._logFile = new temporary.File();
     this._logStream = fs.createWriteStream(this._logFile.path);
     task.stream.pipe(this._logStream);
-    task.stream.on('data', () => condition.broadast());
+    task.stream.on('data', () => condition.broadcast());
     task.stream.on('end', () => condition.broadcast());
 
     // add the bulk log
@@ -107,19 +107,25 @@ class WebhookLogs {
     await this._bulkLog.created(task);
 
     // create hook ID and hook to add to the WebhookServer
-    let hookID = slugid.nice() + slugid.nice();
+    let hookID = slugid.nice();
+    let self = this;
     let hook = async function(req, res) { 
-      let reader = new ContinuousReader(this._logFile.path);
+      let reader = new ContinuousReader(self._logFile.path);
       try{
         let data = "";
-        while(task.state === 'running' && !res.finished){
+        // write CORS header
+        res.writeHead(200,{
+          'Content-Type': 'text/plain',
+          'Access-Control-Allow-Origin': '*'
+        });
+        while(task.taskState === 'running' && !res.finished){
           data = await reader.readChunk();
           if(data.length > 0) {
             res.write(data);
           }
           await condition.wait();
         }  
-        await reader.pipeUntilEOF(res);
+        await reader.pipeUntilEof(res);
         // write any remaining data;
       }finally{
         res.end();
@@ -171,7 +177,7 @@ class WebhookLogs {
           storageType: 'reference',
           expires: expiration,
           contentType: 'text/plain',
-          url: this.publicUrl
+          url: backingUrl
         }
       );
     }finally{
--- a/src/lib/features/webhooklogs.js
+++ b/src/lib/features/webhooklogs.js
@@ -10,7 +10,7 @@ const debug = Debug('taskcluster-docker-worker:features:webhooklog')
 
 /**
  * ContinuousReader is a utility class which allows reading a file
- * which is continuously being written to. A single call to `readUntilEOF`
+ * which is continuously being written to. A single call to `readChunk`
  * will read the file until EOF. Subsequent calls will continue reading 
  * from the point where the previous read ended.
  */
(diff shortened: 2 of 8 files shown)


src/bin/worker.js
```diff
@@ -239,6 +240,17 @@ program.parse(process.argv);
 
   await taskListener.connect();
 
+  // Set up the webhookServer instance for runtime.
+  var webhookServer = null;
+  try{
+    webhookServer = new WebhookServer(config.taskcluster);
+    let url = await webhookServer.start();
+    debug("webhookserver connected. hosted on "+url);
+  }catch(e){
```
jonasfj: nit: spacing

src/bin/worker.js
```diff
@@ -239,6 +240,17 @@ program.parse(process.argv);
 
   await taskListener.connect();
 
+  // Set up the webhookServer instance for runtime.
+  var webhookServer = null;
+  try{
+    webhookServer = new WebhookServer(config.taskcluster);
+    let url = await webhookServer.start();
+    debug("webhookserver connected. hosted on "+url);
```
jonasfj: Don't use `+` in `debug` statements... using `debug('bla bla: %s', url);` that way when `debug = function(){}` (empty function) which is the case when `DEBUG=''` it won't cost anything...

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+   * The whclient binary must be in bin-utils.
+   * Default target port is 61022.
+   *
+   * @param {String} binPath - Location of whclient binary
+   * @param {String} clientID - Taskcluster client ID
+   * @param {String} accessToken - Taskcluster access token
+   * @param {String} targetPort - Port to which connections should
+   * forwared
+   */
+  constructor({clientId, accessToken}, targetPort){
```
jonasfj: temporary credentials include a `certificate`, just pass creds around as an object and you'll be fine...

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+    this.clientId = clientId;
+    this.accessToken = accessToken;
+    this.targetPort = targetPort;
+    if(!clientId || !accessToken){
+      throw "clientId and accessToken are required for WebhookServer";
+    }
+    if(targetPort == null)
+      this.targetPort = DEFAULT_TARGET_PORT;
```
jonasfj: missing brackets, (nit style thing)

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * @example
+   *
+   * @return {String} - Tunnel endpoint on which worker is hosted
+   */
+  async start() {
+    // start the whclient binary
+    let startClient = new Promise((resolve, reject) => {
+      this.proc = spawn(this.path, 
+        [this.clientId, this.accessToken, this.targetPort],
```
jonasfj: How about passing these as env vars? `targetPort` is a great argument, but checkout tc-client for what env var names we usually use for credentials..

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+        {
+          stdio: ['ignore', 'pipe', 'pipe'],
+        });
+
+      proc.stdout.on('data', (data) => {
```
jonasfj: Won't work... You need to read stdout to end... You can't be sure everything will arrive in one chunk (that's implementation specific)

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+    if(targetPort == null)
+      this.targetPort = DEFAULT_TARGET_PORT;
+
+    this.hooks = {}
+
+    // create a server for handling hooks
+    this.server = http.createServer(this._handler)
+    this.baseURL = '';
+    this.path = path.join(__dirname, "../../bin-utils/whclient")
```
jonasfj: use a nice name for this variable... `whclientPath`

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+            return reject('no url in response');
+          }
+          return resolve(resp.url);
+        }catch(e){
+          return reject(e);
+        }
+      });
+
+      proc.stderr.on('data', (data) => {
+        return reject(data);
```
jonasfj: Wrap in an error object... so we get a stack...

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+          if (resp.url === undefined) {
+            return reject('no url in response');
+          }
+          return resolve(resp.url);
+        }catch(e){
+          return reject(e);
+        }
+      });
+
+      proc.stderr.on('data', (data) => {
```
jonasfj: again, data could be just one chunk...

Also what you are waiting for is:
 1) stdout to be closed (and output containing a url)
 2) process exiting non-zero with some output in stderr

jonasfj: If you also use stdout for logging from `whclient` which makes sense... maybe write the url to a file instead...
you can pass a filename by command line argument, and you can use file system watchers to monitor for creation of the file...

Note: when creating the file in whclient, you would have to create it and then atomically rename it... to avoid docker-worker creating an empty file as soon as it's created...

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+    // req url will start with '/'
+    if (req.url.length() < 24 || req.url[23] !== '/') {
+      //404
+      return
+    }
+    let id = req.url.slice(1, 23);
+    let path = req.url.slice(23);
+    let hook = this.hooks[id];
+    if (hook === undefined) {
+      // 404
```
jonasfj: Send a response of 404...

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+        task.runId,
+        this._logsLocation.live,
+        {
+          storageType: 'reference',
+          expires: expiration,
+          contentType: 'text/plain',
+          url: this.publicUrl
+        }
+      );
+    }finally{
```
jonasfj: nit spacing...

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+ */
+class Signal {
+  constructor() {
+    this._resolve = null;
+    this._promise = new Promise(resolve => this._resolve = resolve);
+  }
+
+  send() {
+    this._resolve();
+    this._promise = new Promise(resolve => this._resolve = resolve);
```
jonasfj: wrong order... calling `_resolve` could if any of them were synchronous cause a new call to `recv` which would then get the old promise...

You need to swap, create new and then signal..

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+ * is called, the promise is returned. Multiple callers can await the same
+ * promise.
+ */
+class Signal {
+  constructor() {
+    this._resolve = null;
+    this._promise = new Promise(resolve => this._resolve = resolve);
+  }
+
+  send() {
```
jonasfj: Let's call this broadcast

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Signal is a utility class which is used for notifying readers when data
+ * has been written to the file. It uses a single promise. The constructor
+ * initializes the `_promise` and `_resolve` fields. When `send` is called,
+ * the previous promise is resolved and the fields are reset. When `recv`
+ * is called, the promise is returned. Multiple callers can await the same
+ * promise.
+ */
+class Signal {
```
jonasfj: Maybe call this `condition`, it's similar to cond in `sync.Cond` (golang) and conditions in pthreads, right?

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+    this._resolve = null;
+    this._promise = new Promise(resolve => this._resolve = resolve);
+  }
+
+  send() {
+    this._resolve();
+    this._promise = new Promise(resolve => this._resolve = resolve);
+  }
+
+  recv() {
```
jonasfj: Perhaps call this `wait`

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+    await this._bulkLog.created(task);
+
+    // create hook ID and hook to add to the WebhookServer
+    let hookID = slugid.nice() + slugid.nice();
+    let hook = async function(req, res) { 
+      let reader = new ContinuousReader(this._logFile.path);
+      try{
+        while(task.state === 'running' && !res.finished){
+          let data = await reader.readUntilEOF();
+          res.write(data);
```
jonasfj: Reading everything to a buffer seems like a really bad idea...

src/lib/features/webhooklogs.js
```diff
@@ -0,0 +1,181 @@
...
+          await sig.recv();
+        }  
+      }finally{
+        res.end();
```
jonasfj: close the reader

src/lib/features/webhooklogs.js
jonasfj: What you want is `getChunkUntilEOF` or something... or `pipe` until EOF... or maybe just `pipe(... {close: false})`

and `autoClose: false` in the reader...

src/bin/worker.js
gregarndt: I'm curious, why does concatenation have a cost, but string interpolation doesn't?

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+   * forwared
+   */
+  constructor({clientId, accessToken}, targetPort){
+    this.clientId = clientId;
+    this.accessToken = accessToken;
+    this.targetPort = targetPort;
+    if(!clientId || !accessToken){
+      throw "clientId and accessToken are required for WebhookServer";
+    }
+    if(targetPort == null)
```
gregarndt: you can default this in the constructor params.

`constructor({clientId, accessToken}, targetPort = DEFAULT_TARGET_PORT)`

src/lib/webhookserver.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  /**
+   * Shutdown the whclient binary and exit
+   */
+  kill() {
+    // send SIGKILL. Whclient will wait for connections to end
+    // for 5 seconds before closing.
+    this.proc.kill();
```
gregarndt: whclient captures SIGKILL and will try to do things before terminating?

jonasfj:
@gregarndt 
> On a different note, it seems that all instances will need to have one port opened up that only the webhook tunnel can speak to. Are we deploying one webhook tunnel server per vpc, per region?

No, `whclient` will run on the worker and forward all incoming requests (arriving from webhooktunnel) to port on localhost... Similar to how ngrok/localtunnel works... This just means we avoid implementing the whclient in node.. The port is not exposed, it's only used for inter-process communication between processes on the worker @ckousik right?

if we wanted to we could probably make garndt feel even better by using a unix domain socket instead of a port... but we can't do unix domain sockets on platforms other than linux... Not that docker-worker runs on anything else, but who knows...

src/lib/webhookserver.js
ckousik: I've added this as a command line argument to `whclient`.

src/lib/webhookserver.js
ckousik: It will stop accepting new connections, and will wait for 2 seconds for any pending connections to close, before exiting.

djmitche:
@ckousik what's the status of this?  Is it possible to resurrect this and get it merged, or should we abandon it?  We are working on dropping support for docker-worker, so it's not worth a lot of effort..

ckousik:
@djmitche It would be difficult for me to get to it now, so I would like to drop it 🙂

djmitche:
OK, thanks for letting me know :)

PR #305, opened 2017-08-18
Bug 1389719: Rename image in manifest.json. r=jonasfj

walac:
Newer versions of docker-worker contain a manifest.json file with the
image tag. If it is present, repositories file is ignored.

This opens a security vulnerability by creating a malicious image with a
a known image name in manifest.json and poison image cache.

### DIFF
--- a/src/lib/docker/artifact_image.js
+++ b/src/lib/docker/artifact_image.js
@@ -255,6 +255,7 @@ export default class ArtifactImage {
     let filename = path.basename(tarballPath, '.tar');
     let editedTarballPath = path.join(dir, filename + '-edited.tar');
     let extractedPath = path.join(dir, filename);
+    let manifestPath = path.join(extractedPath, 'manifest.json');
 
     let extractStream = tarfs.extract(extractedPath);
     fs.createReadStream(tarballPath).pipe(extractStream);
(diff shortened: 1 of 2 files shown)


jonasfj:
Note, we really should implement this using `tar-stream`, see the section on "modifying existing tarballs":
https://github.com/mafintosh/tar-stream#modifying-existing-tarballs

And we should stream the output from that directly to `docker load` non of this would be particularly hard to do either...

ie. pipe binary to `zstd` take stdout from `zstd` and pass it as input to `tar-stream`, then create a new `tar-stream` using the modify existing pattern and pass output of that `tar-stream` as stdin for `docker load`.

gregarndt:
I already have a patch in the works to extract, rename, and load in place.
I was waiting for wander ot be done with this work first.

src/lib/docker/artifact_image.js
```diff
@@ -280,6 +281,16 @@ export default class ArtifactImage {
     newRepoInfo = JSON.stringify(newRepoInfo);
     fs.writeFileSync(path.join(extractedPath, 'repositories'), newRepoInfo);
 
+    if (fs.existsSync(manifestPath)) {
+      let manifest = JSON.parse(fs.readFileSync(manifestPath));
+      if (manifest.length > 1) {
```
gregarndt: Can we add this same check to the repositories file too.

walac: Done.

PR #310, opened 2017-08-30
Update taskcluster proxy image to 4.0.0

gregarndt:
In addition to bumping version, the command passed into taskcluster proxy has changed to explicitly declare what is the task ID.


petemoore:
Note, this new version was backed out when it landed in production.

PR #313, opened 2017-09-05
Bug 1394358 - revert upgrade to relengapi-proxy 2.1.1 which does not work

petemoore:
Rok is going to be away all September, so let's roll back this change so it doesn't mess up any releases we might need to make before relengapi-proxy is fixed.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -31,7 +31,7 @@ defaults:
     randomizationFactor: 0.25
   features:
     relengAPIProxy:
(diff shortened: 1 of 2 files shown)


petemoore:
Note, it might be worth deploying after this is merged, so that master remains a reflection of production (and we are sure the revert worked ok).

gregarndt:
This was already done in PR https://github.com/taskcluster/docker-worker/pull/312 which was awaiting review, but I can just approve this one and close the other.

PR #315, opened 2017-09-12
Bug 1307493: Apply exponential back off to sidecar startup. r=garndt

walac:
sidecar containers are the containers that start together with the task
container and that implement the task features. The service implemented
inside the sidecar cotainer may fail due to network hiccups, causing the
sidecar containers to fail to startup.

By applying startup retry with exponential back off we hope to fix these
random startup fails. Given the low probability of an startup fail, we
retry only 3 times.

### DIFF
--- a/src/lib/task.js
+++ b/src/lib/task.js
@@ -12,6 +12,7 @@ import fs from "mz/fs";
 import child_process from "mz/child_process";
 import taskcluster from 'taskcluster-client';
 import base from 'taskcluster-base';
+import promiseRetry from 'promise-retry';
 
 import features from './features';
 import getHostname from './util/hostname';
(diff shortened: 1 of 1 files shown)


src/lib/task.js
```diff
@@ -12,6 +12,7 @@ import fs from "mz/fs";
 import child_process from "mz/child_process";
 import taskcluster from 'taskcluster-client';
 import base from 'taskcluster-base';
+import promiseRetry from 'promise-retry';
```
gregarndt: Wow, I didn't know about this library, but that certainly makes things easier!  @imbstack have you ever dealt with this library before? I'm not aware of the nuances (if any).

src/lib/task.js
```diff
@@ -816,7 +817,13 @@ export class Task extends EventEmitter {
       // Build the list of container links... and base environment variables
       linkInfo = await this.states.link(this);
       // Hooks prior to running the task.
-      await this.states.created(this);
+      await promiseRetry(retry => {
+        return this.states.created(this).catch(retry);
```
gregarndt: Do all of our state handlers implement a created() method that can be retried?  we should audit our features to ensure that this won't introduce some other problems.

src/lib/task.js
walac: It is used in artifact upload retries.

imbstack: Interesting! And nope, I've never used it before.

PR #316, opened 2017-09-20
Deployment improvements

walac:
A bunch of improvements for AMI building, towards deployment automation. Look at individual commits for details.

### DIFF
--- a/deploy/bin/build
+++ b/deploy/bin/build
@@ -30,8 +30,21 @@ packer_config() {
 }
 
 cmd_app() {
+  local docker_worker_tgz=/tmp/docker-worker-$$.tgz
   # Package up the node app.
-  tar --exclude='./node_modules' --exclude='./*.pub' --exclude='./build/' --exclude='./deploy' --exclude='.test' --exclude='.git' --exclude='.vagrant' --exclude='./.DS_Store' --exclude='./vagrant/' --exclude='app.box' -zcvf docker-worker.tgz .
+  tar \
+    --exclude='./node_modules' \
+    --exclude='./*.pub' \
+    --exclude='./build/' \
+    --exclude='./deploy' \
+    --exclude='.test' \
+    --exclude='.git' \
+    --exclude='.vagrant' \
+    --exclude='./.DS_Store' \
+    --exclude='./vagrant/' \
+    --exclude='app.box' \
+    -zcvf $docker_worker_tgz .
+  mv $docker_worker_tgz docker-worker.tgz
   local git_rev=$(git rev-parse --verify HEAD)
 
   # Configure the deployment if needed...
--- a/deploy/bin/build
+++ b/deploy/bin/build
@@ -81,6 +81,8 @@ then
   exit 1
 fi
 
+rm -f packer-artifacts.json
+
 case "$1" in
 "app")
   cmd_app ${@:2}
@@ -92,3 +94,20 @@ case "$1" in
   cmd_help
   ;;
 esac
+
+# extract only the AMI ids and store in a new json file
+if [ -f packer-artifacts.json ]; then
+  jq '
+    .builds[]
+    | {
+      (.name): [
+        .artifact_id
+        | split(",")
+        | .[]
+        | split(":")
+      ]
+      | flatten
+      | { (.[0]): .[1], (.[2]): .[3], (.[4]): .[5] }
+    }
+  ' packer-artifacts.json | jq -s add > docker-worker-amis.json
+fi
--- a/deploy/bin/gen-signing-key
+++ b/deploy/bin/gen-signing-key
@@ -1,0 +1,36 @@
+#!/bin/bash -e
+
+name=Docker-Worker
+
+# rngd is linux only
+if [ $(uname -s) == Linux ]; then
+  # Generate enough entropy to allow for gpg key generation
+  sudo rngd -r /dev/urandom
+fi
+
+# Generate gpg key
+cat >gen_key_conf <<EOF
+  %echo Generating GPG signing key
+  Key-Type: RSA
+  Key-Length: 2048
+  Name-Real: $name
+  Name-Email: taskcluster-accounts+gpgsigning@mozilla.com
+  %commit
+  %echo Done generating key
+EOF
+
+echo "Generating signing key"
+gpg --batch --gen-key gen_key_conf
+rm gen_key_conf
+
+key=$(gpg --list-key $name | head -1 | awk '{print $2}' | awk -F/ '{print $2}')
+fingerprint=$(gpg --fingerprint $key | grep "fingerprint" | awk -F= '{print $2}' | tr -d " ")
+
+echo "Exporting private signing key"
+gpg -a --export-secret-key $key > docker-worker-gpg-signing-key.key
+
+echo "Exporting public signing key"
+gpg -a --export $key > docker-worker.pub
+
+gpg --batch --delete-secret-and-public-key $fingerprint
+
--- a/deploy/bin/import-docker-worker-secrets
+++ b/deploy/bin/import-docker-worker-secrets
@@ -58,13 +58,17 @@ pass show tls/star.taskcluster-worker.net | awk '
 ' > $base_dir/docker-worker-cert.key
 chmod 0600 $base_dir/docker-worker-cert.key
 
+pass show docker-worker/cot-signing-key > $base_dir/docker-worker-cot-signing.key
+chmod 0600 $base_dir/docker-worker-cot-signing.key
+
 echo '
 {
   "debug.level": "",
   "privateKeyLocation": "'$base_dir/docker-worker.key'",
   "filesystem": "aufs",
   "sslCertificateLocation": "'$base_dir/docker-worker.crt'",
-  "sslKeyLocation": "'$base_dir/docker-worker-cert.key'"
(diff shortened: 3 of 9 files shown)


deploy/bin/import-docker-worker-secrets
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
gregarndt: I would like to get another set of eyes on this because of the security implications and also the use of password store (cc @djmitche) 

What concerns me here is that we are dropping secrets into this directory.  Wander said he would look into moving them elsewhere, but we also want to remove them afterwards so we do not have secrets floating around on people's machines.

Is there a way we could get these secrets from password store, and use them with the AMI generation, without having them dropped unencrypted on disk?

deploy/packer/app.json
```diff
@@ -43,7 +43,7 @@
       "type":           "file",
       "direction":      "download",
       "source":          "/tmp/docker-worker.pub",
-      "destination":    "docker-worker-{{build_name}}-{{timestamp}}.pub"
+      "destination":    "docker-worker-{{build_name}}.pub"
```
gregarndt: the purpose of this was because we were always generating a new private key, so we had to extra the public key from the AMI.  If we are reusing keys, we no longer need this at all.

deploy/packer/app.json
```diff
@@ -79,5 +79,12 @@
         "Base_AMI":         "{{user `pvSourceAMI`}}"
       }
     }
+  ],
+  "post-processors": [
```
gregarndt: I didn't know about this.  I'll have to read up.  pretty cool!

also, `packer build` has a `-machine-readable` that might have helped too.

deploy/packer/app.json
```diff
@@ -26,6 +26,30 @@
...
+    {
+      "type":           "shell",
+      "inline":         ["/tmp/gen-signing-key"],
+      "only":           ["hvm-builder", "pv-builder"]
+    },
+    {
+      "type":           "file",
+      "source":         "{{user `cotSigningKey`}}",
+      "destination":    "docker-worker-gpg-signing-key.key",
+      "only":           ["hvm-builder-trusted"]
```
gregarndt: ++

deploy/bin/import-docker-worker-secrets
walac: We can remove them at the end of the build script, I didn't do it because it was not done before and I thought there was a good reason for that. Btw, I changed the patch to put those into `/tmp`.

deploy/packer/app.json
walac: I removed it in a later commit, if you look at the file changes tab, you will see this code section removed.

deploy/bin/build
```diff
@@ -30,8 +30,21 @@ packer_config() {
 }
 
 cmd_app() {
+  local docker_worker_tgz=/tmp/docker-worker-$$.tgz
   # Package up the node app.
-  tar --exclude='./node_modules' --exclude='./*.pub' --exclude='./build/' --exclude='./deploy' --exclude='.test' --exclude='.git' --exclude='.vagrant' --exclude='./.DS_Store' --exclude='./vagrant/' --exclude='app.box' -zcvf docker-worker.tgz .
+  tar \
+    --exclude='./node_modules' \
```
djmitche: why not just `--exclude=./docker-worker.tgz`?

djmitche:
Oh, that was only looking at one commit, sorry

deploy/bin/import-docker-worker-secrets
djmitche: So one way around this would be to use `/dev/shm` for the sensitive files -- and be careful to put them in an 0700 directory and remove them afterward.

deploy/bin/import-docker-worker-secrets
```diff
@@ -0,0 +1,67 @@
...
+# This script reads docker-worker store in passwordstore and writes
+# up deploy.json.
+#
+# Use: ./import-docker-worker-secrets --passphrase=<your-passphrase>
+#
+# Notice you must have taskcluster passwordstore cloned and configured.
+# See ssh://gitolite3@git-internal.mozilla.org/taskcluster/secrets.git
+# for details.
+
+export PASSWORD_STORE_GPG_OPTS="$*"
```
djmitche: This probably deserves a header comment too?  Or maybe just generally the user is expected to pass in any necessary PASSWORD_STORE_.. env variables to make `pass` work? (for example, I need to set `PASSWORD_STORE_DIR`)

deploy/bin/import-docker-worker-secrets
```diff
@@ -0,0 +1,67 @@
...
+
+base_dir=/tmp
+
+pass show aws/workers-key > $base_dir/docker-worker.key
```
djmitche: Is `set -e` enabled here?  If this (or the later pass invocations) fails, the script should fail..

deploy/bin/import-docker-worker-secrets
```diff
@@ -0,0 +1,67 @@
...
+#
+# Use: ./import-docker-worker-secrets --passphrase=<your-passphrase>
+#
+# Notice you must have taskcluster passwordstore cloned and configured.
+# See ssh://gitolite3@git-internal.mozilla.org/taskcluster/secrets.git
+# for details.
+
+export PASSWORD_STORE_GPG_OPTS="$*"
+
+base_dir=/tmp
```
djmitche: setting this to `mktemp -d /dev/shm/docker-worker-XXXXX` would make me feel a lot better :)

deploy/bin/build
```diff
@@ -92,3 +94,20 @@ case "$1" in
...
+      (.name): [
+        .artifact_id
+        | split(",")
+        | .[]
+        | split(":")
+      ]
+      | flatten
+      | { (.[0]): .[1], (.[2]): .[3], (.[4]): .[5] }
+    }
+  ' packer-artifacts.json | jq -s add > docker-worker-deploy.json
```
djmitche: that's some serious jq ninjery there.

deploy/bin/gen-signing-key
```diff
@@ -0,0 +1,36 @@
...
+EOF
+
+echo "Generating public signing key"
+gpg --batch --gen-key gen_key_conf
+rm gen_key_conf
+
+key=$(gpg --list-key $name | head -1 | awk '{print $2}' | awk -F/ '{print $2}')
+fingerprint=$(gpg --fingerprint $key | grep "fingerprint" | awk -F= '{print $2}' | tr -d " ")
+
+echo "Exporting public signing key"
```
djmitche: s/public/private/

deploy/bin/gen-signing-key
```diff
@@ -0,0 +1,36 @@
...
+  %echo Generating GPG signing key
+  Key-Type: RSA
+  Key-Length: 2048
+  Name-Real: $name
+  Name-Email: taskcluster-accounts+gpgsigning@mozilla.com
+  %commit
+  %echo Done generating key
+EOF
+
+echo "Generating public signing key"
```
djmitche: s/public //

(it really generates both, right?)

deploy/bin/build
walac: That was my first try, but if the file doesn't exist beforehand, tar ignores the `--exclude` option.

djmitche: Oh right.. *recalls vague memories from work at Zmanda*

deploy/packer/app.json
```diff
@@ -26,6 +26,30 @@
...
+      "inline":         ["chmod u+x /tmp/gen-signing-key"]
+    },
+    {
+      "type":           "shell",
+      "inline":         ["/tmp/gen-signing-key"],
+      "only":           ["hvm-builder", "pv-builder"]
+    },
+    {
+      "type":           "file",
+      "source":         "{{user `cotSigningKey`}}",
```
jonasfj: Are generating the private key on our host/master machine and then copying it to the AMI?

Note: I think it would be slightly safer to generate the key on the AMI and copy out the public key, that way it ever only exists on the worker AMI, the private key never gets into memory or disk of the host/master machine...

Granted this might be a follow up someday when we have time :)

djmitche: We're using the same private/public keypair on all AMIs.

deploy/bin/import-docker-worker-secrets
walac: This is intentional to use the command line to pass arguments to gpg using the command line. In special, the `--passphrase` option so the user doesn't need to type the passphrase each time `pass` is called.

walac: Good catch on the `chmod`, I added it. The intention of this script is just to automate the generation of `deploy.json` file, which has been done manually so far. As we will soon move to automated deployment, I rather prefer not to invest much more time on this as it will be obsolete soon.

deploy/bin/import-docker-worker-secrets
walac: The problem is that this is not portable. macOS, for example, doesn't implement /dev/shm.

deploy/bin/import-docker-worker-secrets
walac: Fixed.

walac:
I think I fixed all pointed issues. Thank you all for the valuable code review :)

PR #320, opened 2017-09-25
[deploy] Add scripts to update provisioner. r=garndt

walac:
We add two new utility scripts to help deployment:

* update-worker-types: update the aws-provisioner with the new AMI ids.
It produces a backup of provisioner configuration before updating.
* rollback-worker-types: used to rollback the new AMIs in case of
deployment bustage.

Both scripts kill worker-types running EC2 instances.


walac:
Adding @eliperelman to review the quality of JS code.

deploy/bin/rollback-worker-types.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
+
+const utils = require('./worker_type_utils');
```
gregarndt: I believe `import foo from 'bar';` is the way of the future for imports.  Eli can correct me if I'm wrong.

deploy/bin/update-worker-types.js
```diff
@@ -0,0 +1,56 @@
...
+const jsonfile = bluebird.Promise.promisifyAll(require('jsonfile'));
+const utils = require('./worker_type_utils');
+
+/*
+ * Write worker-types configuration to a json file
+ */
+async function backupWorkerTypes(client) {
+  console.log('Creating worker-types backup file.');
+  const names = await client.listWorkerTypes();
+  const workerTypes = names.map(name => {
```
gregarndt: to avoid confusion, could this be named differently than the JSON file you imported?

deploy/bin/update-worker-types.js
```diff
@@ -0,0 +1,56 @@
...
+    const w = new utils.WorkerType(client, name);
+    return w.workerType();
+  });
+
+  await Promise.all(workerTypes).then(
```
gregarndt: The Promise.all() might read better when initalizing workerTypes than here.  Took me a second to realize that w.workerType() is actually a promise.

deploy/bin/update-worker-types.js
```diff
@@ -0,0 +1,56 @@
...
+    result => jsonfile.writeFileAsync(
```
gregarndt: we only use jsonfile once here.  Is it worth using it and needing another module for this one use case?

gregarndt: Also since we're only writing one file here, why not use writeFileSync and not need bluebird to promisfy it?

deploy/bin/update-worker-types.js
```diff
@@ -0,0 +1,56 @@
...
+    return workerTypes[i].map(async name => {
+      const workerType = new utils.WorkerType(client, name);
+      const wt = await workerType.workerType();
+
+      for (region of wt.regions) {
+        region.launchSpec.ImageId = amis[i][region.region];
+      }
+      console.log(`Updating ${name}`);
+      return workerType.update(wt).then(
+        () => utils.killInstances(workerType)
```
gregarndt: I do not feel comfortable with this.  This would be mean that when we deploy a new ami, it'll kill instances currently running a task.  That means many thousands of instances (and jobs failed with claim expired).  Perhaps we can adapt something like what generic-worker is doing and have a deployment id generated.  In between claims of tasks, if the deployment ID has changed from when the worker started, it will shut down.

deploy/bin/worker_type_utils.js
```diff
@@ -0,0 +1,69 @@
+const tc = require('taskcluster-client');
+const promiseRetry = require('promise-retry');
```
gregarndt: I'm not sure the use of this library here.  The taskcluster client already has retries in it.

deploy/bin/update-worker-types.js
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env node
+
+const bluebird = require('bluebird');
```
gregarndt: if we do end up using bluebird in the end, I noticed it was not included in package.json.  I would like to first address the need for this library as well as promisfying jsonfile that I mentioned in my other comment.

deploy/bin/rollback-worker-types.js
walac: I am using this syntax to be node 8 compatible.

deploy/bin/update-worker-types.js
walac: I am using bluebird only because it was already present in `package.json`.

deploy/bin/update-worker-types.js
walac: Hrm, I think that expression is already too complex, tbh.

deploy/bin/update-worker-types.js
walac: bluebird is already present in docker-worker, the as `backupWorkerTypes` already returns a `Promise`, using an async version seems the right thing.

deploy/bin/update-worker-types.js
walac: My understanding was that we should kill all instances when deploying new AMIs. If this is not the case, I will remove it. I need to check the generic-worker strategy.

deploy/bin/worker_type_utils.js
walac: When using [manage.js](https://github.com/taskcluster/aws-provisioner/blob/master/src/manage.js) I had some issues.

deploy/bin/rollback-worker-types.js
gregarndt: hrm, perhaps that's the problem.  docker-worker is node 6 compatible but we are introducing node 8 for the management/deployment side of things.  Are we also moving the worker code base to be node 8 compatible?

deploy/bin/update-worker-types.js
gregarndt: hrm, I don't see it here: https://github.com/taskcluster/docker-worker/blob/master/package.json

deploy/bin/worker_type_utils.js
gregarndt: do you remember the issues? These sound like something we need to open up against aws-provisioner then.

deploy/bin/update-worker-types.js
gregarndt: fair enough. :)

gregarndt:
Thanks for making this deployment and management so much better!

deploy/bin/rollback-worker-types.js
walac: That's the plan

deploy/bin/worker_type_utils.js
walac: When doing batch worker-type updates, sometimes one randomly failed. Adding `promiseRetry` fixed it.

deploy/bin/update-worker-types.js
walac: I renamed the imported json file.

deploy/bin/update-worker-types.js
walac: Err, you are right, I must have fooled myself look at one of the `node_modules/*` `package.json`.

gregarndt:
Approved as long as we are not killing instances when deploying new amis and only killing instances using the ami we need to roll away from.

PR #321, opened 2017-09-28
Pin yarn to 1.0.2

gregarndt:
There is an issue with the latest version of yarn, version 1.1.0 where
devDependencies are installed unintentionally.  This happens even when
NODE_ENV=production or passing in `--production` to yarn.  This causes
dockerode-promise to install an older version of dockerode which does
not work with image tags containing a content hash, such as
some-image@sha256:1234.


gregarndt:
This is somehow probably related to: https://github.com/yarnpkg/yarn/issues/4557

eliperelman:
https://github.com/yarnpkg/yarn/issues/4557

djmitche:
Does that version of dockerode-promise need to be updated to correspond with the other dockerodes?

djmitche:
also,

WAT

PR #322, opened 2017-09-28
Automated deployment

walac:
Several small fixes for deployment scripts. See the commit messages for details.

### DIFF
--- a/deploy/bin/rollback-worker-types.js
+++ b/deploy/bin/rollback-worker-types.js
@@ -2,15 +2,24 @@
 
 const utils = require('./worker_type_utils');
 const backup = require('../../worker-types-backup.json');
+const _ = require('lodash');
+
+async function updateWorkerType(workerType, regions) {
+  const definition = await workerType.workerType();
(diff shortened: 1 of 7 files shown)


deploy/bin/rollback-worker-types.js
```diff
@@ -2,15 +2,24 @@
 
 const utils = require('./worker_type_utils');
 const backup = require('../../worker-types-backup.json');
+const _ = require('lodash');
+
+async function updateWorkerType(workerType, regions) {
+  const definition = await workerType.workerType();
+  definition.regions = regions;
+  await workerType.update(definition);
+  await utils.killInstances(workerType);
```
gregarndt: is killinstances updated to only kill instances which are running the ami you are rolling back?  We do not want to kill all instances for a worker type because it's quite possible there are good machines out there running the older AMI already.

gregarndt: Nevermind, I misread the other method update, I see you did it.

PR #325, opened 2017-10-05
Bug 1405889 - Upgrade taskcluster-proxy from 4.0.0 to 4.0.1

petemoore:
I've [released](https://hub.docker.com/r/taskcluster/taskcluster-proxy/tags/) taskcluster-proxy 4.0.1 with the [bug 1405889](https://bugzilla.mozilla.org/show_bug.cgi?id=1405889) taskcluster-client-go [fix](https://github.com/taskcluster/taskcluster-client-go/commit/c16489a70a6f20ac72964be79df37753b3fce4d5).

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -53,7 +53,7 @@ defaults:
   restrictCPU: false
 
   # Image used to  create the taskcluster proxy container.
-  taskclusterProxyImage: 'taskcluster/taskcluster-proxy:4.0.0'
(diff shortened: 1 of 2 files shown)


gregarndt:
Reviewed the test failures, and one is a known CI failure (doesn't fail locally) and the other two are odd, but in no way use the proxy so this change couldn't have done it.

PR #326, opened 2017-10-10
Add Balrog stage host/ip to balrogvpnproxy hack.

mozbhearsum:
We're going to start using the Balrog stage environment more regularly for staging Releases. This looks like one of the things we'll need to make sure Funsize is able to submit to it.

### DIFF
--- a/src/lib/features/balrog_vpn_proxy.js
+++ b/src/lib/features/balrog_vpn_proxy.js
@@ -77,7 +77,7 @@ export default class BalrogVPNProxy {
         ExtraHosts: [
           // XXX: hack for now.  Problem in taskcluster-vpn-proxy where resolv.conf
           // isn't updated when vpn'ed in so name resolution does not work.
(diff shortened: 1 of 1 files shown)


src/lib/features/balrog_vpn_proxy.js
```diff
@@ -78,6 +78,7 @@ export default class BalrogVPNProxy {
           // XXX: hack for now.  Problem in taskcluster-vpn-proxy where resolv.conf
           // isn't updated when vpn'ed in so name resolution does not work.
           "aus4-admin.mozilla.org:52.26.16.60"
```
walac: We need to add a comma here, don't we?

mozbhearsum: Whoops, of course!

mozbhearsum:
After digging some more, I realized this won't do what I want...simply adding an extra DNS entry isn't going to make it possibly for the balrogVpnProxy to work with both Balrog stage and prod.

PR #328, opened 2017-10-11
Add VPN proxy for balrog stage

mozbhearsum:
As discussed on IRC, this is a clone of the production proxy. It uses a separate feature name ("balrogStageVPNProxy") and a different host/ip in the feature implementation, but is otherwise the same as the existing one.

I tried to run tests locally for this but I got a bunch of failures that didn't seem related to my patch (eg: image manager failures), so I didn't pursue that further.


gregarndt:
This looks to be what is advertised...a copy/paste of the original proxy with some slight modifications.  Thanks for updating the docs too! we'll need to make sure that whatever has scopes to use the proxy now also gets this scope.

gregarndt:
Two failures in the tests are known and shouldn't be caused by this PR

PR #330, opened 2017-10-13
Introduce `worker:cache:<name>` scope

jonasfj:
We could talk about doing the same for permissions and other stuff...

Note: this is more of a suggestion, we should probably not merge until we have some idea of how we want do this with other scopes... and maybe a plan for doing matching changes in-tree too..

### DIFF
--- a/src/lib/task.js
+++ b/src/lib/task.js
@@ -94,10 +94,13 @@ Create a list of cached volumes that will be mounted within the docker container
 @param {object} volumes to mount in the container
  */
 async function buildVolumeBindings(taskVolumeBindings, volumeCache, taskScopes) {
-  let allowed = await hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes);
+  let allowed = (
+    (await hasPrefixedScopes('worker:cache:', taskVolumeBindings, taskScopes)) ||
+    (await hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes))
+  );
   if (!allowed) {
     throw new Error('Insufficient scopes to attach cache volumes.  The task must ' +
-    'have scope `docker-worker:cache:<cache-name>` for each cache in `payload.caches`.');
+    'have scope `worker:cache:<cache-name>` for each cache in `payload.caches`.');
   }
 
   let bindings = [];


gregarndt:
Definitely would allow for us to support both types of scopes while waiting for all automation to catch up.  

any value in having the worker log when a legacy scope is encountered and then we can audit it easily later?

I think we should update the tests to test for the new scope rather than the legacy one (or both for the time being)

jonasfj:
> any value in having the worker log when a legacy scope is encountered and then we can audit it easily later?

yes, that seems like a neat little 2 second hack :)

> I think we should update the tests to test for the new scope rather than the legacy one (or both for the time being)

Agreed..

It's probably worth doing a grep of the source too :) hehe

src/lib/task.js
```diff
@@ -94,10 +94,13 @@ Create a list of cached volumes that will be mounted within the docker container
 @param {object} volumes to mount in the container
  */
 async function buildVolumeBindings(taskVolumeBindings, volumeCache, taskScopes) {
-  let allowed = await hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes);
+  let allowed = (
+    (await hasPrefixedScopes('worker:cache:', taskVolumeBindings, taskScopes)) ||
+    (await hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes))
```
walac: I think this is better:
```javascript
const allowed = await Promise.all(
  hasPrefixedScopes('worker:cache:', taskVolumeBindings, taskScopes)),
  hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes))
).then(allowed => allowed.some());
```

src/lib/task.js
```diff
@@ -94,10 +94,13 @@ Create a list of cached volumes that will be mounted within the docker container
...
+  );
```
walac: I think this is better:
```javascript
const allowed = await Promise.all(
  hasPrefixedScopes('worker:cache:', taskVolumeBindings, taskScopes)),
  hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes))
).then(allowed => allowed.some());
```

gregarndt: is some() a new method on promises?

walac: `Promise.all` yields an array with the returned result of each async operation. [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is a new Array method in ES6.

gregarndt: Oh...I thought Promise.all() returned a single promise back, not an array.

walac: It does, but this new `Promise` when awaited, yields an array with the results of each individual Promise.

jonasfj: or:
```javascript
const allowed = await Promise.all(
  hasPrefixedScopes('worker:cache:', taskVolumeBindings, taskScopes)),
  hasPrefixedScopes('docker-worker:cache:', taskVolumeBindings, taskScopes))
).then(_.some);
```
or
```javascript
const allowed = await Promise.all([
  'worker:cache:', 'docker-worker:cache:',
].map(prefix => hasPrefixedScopes(prefix, taskVolumeBindings, taskScopes))).then(_.some);
```

Regardless, we should probably just fix it so that `hasPrefixedScopes` isn't async, because it doesn't have to be :)

walac: I liked the last one.

PR #331, opened 2017-10-15
Move to node 8 and knock out babel.

walac:
Node 8 has almost everything we use from ES6 and ES7, except by ES6
module syntax.

We also green up tests.


src/bin/worker.js
```diff
@@ -92,6 +92,10 @@ program.parse(process.argv);
 
 // Main.
 (async () => {
+  process.on('unhandledRejection', (reason, p) => {
```
gregarndt: I think that lib-monitor catches this and sends to sentry.  Is this for just more verbose logging of such things?

src/lib/features/docker_save.js
```diff
@@ -32,10 +31,12 @@ export default class DockerSave {
     let image = task.runtime.docker.getImage(`${imageName}:latest`);
     let imgStream = await image.get();
     let zipStream = zlib.createGzip();
-    imgStream.pipe(zipStream).pipe(fs.createWriteStream(pathname));
     await new Promise((accept, reject) => {
-      zipStream.on('end', accept);
+      const output = fs.createWriteStream(pathname);
       zipStream.on('error', reject);
+      zipStream.on('end', accept);
+      output.on('error', reject);
```
gregarndt: \o/

src/bin/worker.js
walac: When running tests locally, I ran into some "unhandled reject" errors and had to add this code to figure them out.

walac:
The current model of tc-github makes it impossible for tests to pass since it reads .taskcluster.yml file from master branch.

gregarndt:
It's reading the .taskcluster.yml from your PR.  See this task that is using the image hash you updated to in this PR:
https://tools.taskcluster.net/groups/Jv45dMcvTSWety-jXyD48w/tasks/Jv45dMcvTSWety-jXyD48w/details

walac:
> It's reading the .taskcluster.yml from your PR. See this task that is using the image hash you updated to in this PR:
> https://tools.taskcluster.net/groups/Jv45dMcvTSWety-jXyD48w/tasks/Jv45dMcvTSWety-jXyD48w/details

You are right. So, why isn't it running the tests on my patch?

cc @imbstack @djmitche

gregarndt:
Looks like it's failing on this:
https://tools.taskcluster.net/groups/Jv45dMcvTSWety-jXyD48w/tasks/Jv45dMcvTSWety-jXyD48w/runs/0/logs/public%2Flogs%2Flive_backing.log#L5347

djmitche:
It looks like it ran in taskcluster/worker-ci@sha256:125150396fbac3b0e8caf5880ebf10fc952bc2357d1509e9a6bee6872838fa5e and checked out master then merged ed436eeb0cb4732c795ae6f72062ef4a6e86a76c then ran `build.sh`.  What's the problem?

walac:
> It looks like it ran in taskcluster/worker-ci@sha256:125150396fbac3b0e8caf5880ebf10fc952bc2357d1509e9a6bee6872838fa5e and checked out master then merged ed436ee then ran build.sh. What's the problem?

If you look at the patch, the ` import assert from 'assert';` line was changed to `const assert = require('assert');`, which indicates that it is not applying the merge correctly. I can't understand why

PR #332, opened 2017-11-03
Remove race conditions from stream piping. r=garndt

walac:
In several parts of the code there is a common pattern to combine nodejs
pipe method and waitForEvent:

stream1.pipe(stream2);
await waitForEvent(stream1, 'end');

It turns out this code is racy and may end up in a deadlock. If the 'end'
event arrives before the 'end' event, waitForEvent will get stuck
forever.

We solve it by using the promisepipe package, which supplies a promised
version of the pipe function and replace the code above with
promisepipe.

Update: this branch evolved and now includes a lot of fix for intermittent test failures.


gregarndt:
This certainly seems a lot nicer, I wasn't aware of promisepipe!

djmitche:
I don't understand streams well enough to say much about this.  But Greg seems happy!

gregarndt:
Appears the tests aren't happy with this.  We'll need to investigate those.

walac:
> Appears the tests aren't happy with this. We'll need to investigate those.

Yeah, I expected that, had to push the PR to run the tests on CI.

jonasfj:
```js
// state1
stream1.pipe(stream2);
// state2
await waitForEvent(stream1, 'end');
// state3
```

No events can fire between state1 and state2. If they do, the events goes on to the event queue and gets fired when the promise is awaited. **NO, I'm wrong**... I realize as I'm writing.... events can fire... you can call all the event handlers synchronously... And then we would miss the event... In that case the fix could have been.

```js
const done = waitForEvent(stream1, 'end');
stream1.pipe(stream2);
await done;
```

But yeah, using this magic `pipe` seems better...

PR #333, opened 2017-11-07
Fix scopes unit test.

walac:
scopes.hasPrefixedScopes is an async function and any call to it must
`await`. For reasons we probably we never know, the code used to pass
until we changed it in [1]. Now it returns a pending promise.

We fix this bad behavior by awaiting hasPrefixedScopes.

[1] https://github.com/taskcluster/docker-worker/pull/332

### DIFF
--- a/test/test_util_scopes.js
+++ b/test/test_util_scopes.js
@@ -11,8 +11,8 @@ suite('test lib/util/scopes', function() {
     ['starScope',   'pfx:', { vol1: '/vol1', vol2: '/vol2' }, ['pfx:*'], true],
     ['noResources', 'pfx:', {}, [], true],
   ].forEach((t)=> {
-    test(t[0], function() {
-        assert.equal(scopes.hasPrefixedScopes(t[1], t[2], t[3]), t[4]);
+    test(t[0], async function() {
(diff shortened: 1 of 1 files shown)


walac:
> I wonder if pipestream changes Promise somehow so that new Promise() != true? At any rate, this is definitely a good fix.

That's a question we are doomed to never know the answser.

test/test_util_scopes.js
```diff
@@ -11,8 +11,8 @@ suite('test lib/util/scopes', function() {
     ['starScope',   'pfx:', { vol1: '/vol1', vol2: '/vol2' }, ['pfx:*'], true],
     ['noResources', 'pfx:', {}, [], true],
   ].forEach((t)=> {
```
jonasfj: @walac, this should probably be a `map` and wrapped in a `await Promise.all(...`
And then `suite(...., async function() {` would probably be a good idea...

djmitche: It's just calling test(..) with a function argument -- it's up to the test runner to wait for the promises.

jonasfj: ah, I'm not even sure it works on the suite level...

PR #335, opened 2017-11-13
[test] Dump pending async IO operations every 5 seconds. r=garndt

walac:
[test] Dump pending async IO operations every single operation. r=garndt

When running docker-worker tests, a subtle bug in a new implementation
might leave a random async IO pending forever. This a hard bug to track
down, as it can be anywhere in the code.

Now, if we run mocha with '--require src/lib/async-dump' parameter,
docker-worker will dump the pending async IO operations every 5 seconds
(hardcoded).

Reference:
https://gist.github.com/boneskull/7fe75b63d613fa940db7ec990a5f5843


test/dockerworker.js
```diff
@@ -71,7 +71,9 @@ class DockerWorker {
           'securityfs',
           '/sys/kernel/security',
           '&&',
-          `node /worker/src/bin/worker.js`,
+          'node',
+          global.asyncDump ? '--require /worker/src/lib/async-dump' : '',
```
gregarndt: This file does not exist.

walac: *facepalm*, fixed...

gregarndt:
Something is odd with the test failing in chunk 5, it failed exactly the same way both times so I'm wondering if this is a legit problem.  Couldn't be because of this patch, but maybe something else? I don't recall seeing this one fail before.

walac:
It passed when I triggered it again. One of those tests what fall into the dark "some times it doesn't work" situations.

gregarndt:
Hrm, ok.  I can't put my finger on it, but it seems that some of these are more intermittent than they used to be but I am not sure of the root cause.

gregarndt:
Approved this PR because I don't think anything is wrong with it, but I do worry that there are some classes of things failing that we don't know why just yet and putting it into production later will be problematic.

PR #337, opened 2017-11-18
Upgrade base AMIs to latest available

indygreg:
These correspond to the ubuntu-trusty-14.04-amd64-server-20171115.1
releases. The HVM AMI is ENA enabled, which means we should be able to
use c5 instances with it.

I'm not able to get packer to work. Hopefully someone can use this
change, build the base AMI, then plug it into app.json to get the
app images updated.


djmitche:
What are those source AMIs?

indygreg:
Read the commit message. You can verify the AMIs by searching for that string in the "Community AMIs" in the AWS web console or at https://cloud-images.ubuntu.com/locator/ec2/.

indygreg:
I need to send some updated patches. The c5 AMIs have their EBS volumes in /dev/nvme*. So the /etc/default/docker scripts need to be updated to look in those.

The good news is I was able to get packer running in my own AWS account and I produced an AMI that successfully runs overlay2 on a c5.4xlarge instance.

djmitche:
Fair enough -- I forgot these are JSON files and thus not amenable to comments.

walac:
I will work on generating the AMIs and updating the files.

indygreg:
I want to emphasize that this is the first time I've touched this code. I ran no tests beyond "launch a base AMI and spot check Docker."

PR #338, opened 2017-11-24
Use docker-worker overlay fs in app image

walac:


walac:
@djmitche I just tested with latest tc-client, and the bug still holds.

djmitche:
Was that 3.1.1?

indygreg:
Are the new app AMIs available? I've been itching to evaluate performance of the c5 instances...

walac:
I am releasing now :)

PR #340, opened 2017-11-29
Warn about lack of Taskcluster credentials

walac:
Now that we must generate Taskcluster credentials on the fly, add a
check to the release script for lack of proper Taskcluster credentials.


jonasfj:
The same for aws vrede...

Den 28. nov. 2017 21.58 skrev "Wander Lairson Costa" <
notifications@github.com>:

PR #344, opened 2017-12-06
Fix volume mounting with overlay2 fs.

walac:
/dev/nvme0 must be excluded from LVM group, as we do in aufs.

@indygreg in my tests this seems to fix the docker daemon startup.

### DIFF
--- a/deploy/template/etc/default/docker-overlay2
+++ b/deploy/template/etc/default/docker-overlay2
@@ -8,7 +8,7 @@ if ! lvdisplay | grep instance_storage; then
   # c5 and newer has nvme* devices. The nvmeN devices can't be used
   # with vgcreate. But nvmeNnN can.
   if [ -e /dev/nvme0 ]; then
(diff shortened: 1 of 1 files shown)


indygreg:
Eh? Shouldn't `ls nvme*n*` not list `nvme1` since it doesn't have an `n`? But the new code isn't harmful. So if it works, ship it.

walac:
> Eh? Shouldn't ls nvme*n* not list nvme1 since it doesn't have an n? But the new code isn't harmful. So if it works, ship it.

The problem is that it can't list nvme0*, because nvme0n1 for some reason is used somewhere... At least that's how it works on aufs.

indygreg:
The current AUFS code is buggy. The initial patch to support /dev/nvme was buggy. When I patched it, I only fixed overlayfs and not the aufs or btrfs files. So don't look at the AUFS init script as an example of correct.

PR #346, opened 2017-12-08
Tweak ext4 mount options for performance

indygreg:
Docker workers are ephemeral. They don't live longer than a single
boot.

The default configuration of Linux makes reasonable attempts to not
lose your data or corrupt your filesystem. It does this by ensuring
filesystem writes occur in a specific order and semi-frequently.

This commit relaxes the mount options of the ext4 volume used to back
Docker. We essentially disable protections around writing ordering
guarantees and lower the interval that ext4 performs background
flushing of pending writes.

With these changes, the filesystem will almost certainly be corrupted
if power loss or unclean shutdown occurs. However, the filesystem
is ephemeral - just like the instance is - so we shouldn't care.

The most contentious change in the lot is likely the change of the
"commit" value. We definitely want to increase this for performance.
However, I'm not sure if we should increase it to "infinity" or leave
it at something reasonable. The danger with increasing to "infinity"
is that a large amount of writes may pile up and when a sync() is
eventually performed or requested by an application, it could take
a very long time. I /think/ there are other flushing mechanisms and
high water marks in Linux that will kick in and mitigate this
scenario. But I didn't want to take the chance. So this change
only increases the commit interval from 5s to 60s. That's still
small, but is 12x longer than the previous value. So hopefully it
has some positive impact.

### DIFF
--- a/deploy/template/etc/default/docker-overlay2
+++ b/deploy/template/etc/default/docker-overlay2
@@ -43,7 +43,32 @@ then
   else
     echo "Succesfully formated 'instance_storage' as ext4."
     echo "Mounting logical volume"
-    mount /dev/instance_storage/all /mnt
+
+    # Our assumption is that workers are ephemeral. If errors are encountered, the
+    # worker should be thrown away. Workers are never rebooted. So filesystem
+    # durability isn't too important to us.
+
+    # Default mount options: rw,relatime,errors=remount-ro,data=ordered
+    #
+    # We make the following changes:
+    #
+    # errors=panic -- The worker is unusable if the mount isn't writable. So
+    # panic if we encounter this.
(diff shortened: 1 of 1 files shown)


walac:
I am not sure why, but some tests seem to fail with this change: https://treeherder.mozilla.org/#/jobs?repo=try&revision=a58f367b0abf&selectedJob=156491090

One of those I switched to another worker-type and it worked https://tools.taskcluster.net/groups/P2LkKl85QSK1q94F2x1tKA/tasks/P2LkKl85QSK1q94F2x1tKA/details

indygreg:
We should ignore this PR for now. I'll get back around to it after the provisioning code is cleaned up a bit.

PR #347, opened 2018-01-02
Add support for purging caches based on task exit status.

tomprince:
This should allow failures like https://bugzilla.mozilla.org/show_bug.cgi?id=1297151 and https://bugzilla.mozilla.org/show_bug.cgi?id=1426445 gracefully.

### DIFF
--- a/schemas/payload.json
+++ b/schemas/payload.json
@@ -219,7 +219,16 @@
             "title": "Exit statuses",
             "type": "number"
           }
-        }
+        },
+	"purgeCaches": {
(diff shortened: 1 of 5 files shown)


djmitche:
If we take this approach, we should do the same in tc-worker..

tomprince:
Note: I've not tested this, as I don't have the setup for it.

djmitche:
@walac I'll leave this to you to test, merge, and/or work on implementing in tc-worker as you see fit.

PR #348, opened 2018-01-03
Use correct error string.

walac:
In Bug 1387365 we see a strange behavior where an error is thrown but
the Error object is underfined. Instead of err.message, we use the Error
object as the string format, so if it is undefined, we still have some
information about the original error.

### DIFF
--- a/src/lib/states.js
+++ b/src/lib/states.js
@@ -50,12 +50,12 @@ class States {
           return hook[method](task)
             .then(info => { return info; })
             .catch(err => {
-              errors.push(new Error(`Error calling '${method}' for ${hook.featureName} : ${err.message}`));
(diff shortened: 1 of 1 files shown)


src/lib/states.js
```diff
@@ -50,12 +50,12 @@ class States {
           return hook[method](task)
             .then(info => { return info; })
             .catch(err => {
-              errors.push(new Error(`Error calling '${method}' for ${hook.featureName} : ${err.message}`));
+              errors.push(new Error(`Error calling '${method}' for ${hook.featureName} : ${err}`));
```
jonasfj: Nice... generally what I do is:
`err.toString()` or `err ? err.message || err : 'unknown'`

Or maybe `(err || 'unknown').toString()` or just `err || 'unknown'`

PR #349, opened 2018-01-08
Update worker types sequentially.

walac:
Concurrently updating worker types may cause aws-provisioner to timeout
requests.

### DIFF
--- a/deploy/bin/update-worker-types.js
+++ b/deploy/bin/update-worker-types.js
@@ -44,10 +44,10 @@ function backupWorkerTypes(client) {
   );
 }
 
-function updateWorkerTypes(client) {
+async function updateWorkerTypes(client) {
   console.log('Updating worker types...');
-  const p = Object.keys(workerTypesList).map(i => {
-    return workerTypesList[i].map(async name => {
+  for (let i of Object.keys(workerTypesList)) {
+    for (let name of workerTypesList[i]) {
       const workerType = new utils.WorkerType(client, name);
       const wt = await workerType.workerType();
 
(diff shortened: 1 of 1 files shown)


deploy/bin/update-worker-types.js
```diff
@@ -44,22 +44,21 @@ function backupWorkerTypes(client) {
   );
 }
 
-function updateWorkerTypes(client) {
+async function updateWorkerTypes(client) {
   console.log('Updating worker types...');
-  const p = Object.keys(workerTypesList).map(i => {
-    return workerTypesList[i].map(async name => {
+  for (i of Object.keys(workerTypesList)) {
+    for (name of workerTypesList[i]) {
```
jonasfj: You are going to want to do `for (let name of ...) {`

When you omit the `let`, `var` or `const` you are referencing global variables... as this is assignment that works... but if two of these were running in parallel.. which can happen when they are async  this could be a LOT of fun, haha :)

Check out:
```
[~]$node
> console.log(i)
ReferenceError: i is not defined
> for (i of [1,2]) {console.log(i)}
1
2
undefined
> console.log(i)
2
undefined
> for (let j of [1,2]) {console.log(j)}
1
2
undefined
> console.log(j)
ReferenceError: j is not defined
```

If you don't declare `name` using something like `let name`, then `name = ...` is the as `global.name = ...` (in node) or `window.name = ...` (in the browser context).

Anyways, fun fact that should scare us into using a linter, or at least `"use strict"` :)

jonasfj: For the record, I've made this mistake a few times myself... and usually works great... until it stops working and then nothing makes sense when you try to debug it.

walac: Good catch, thanks :)

walac:
@jonasfj I need you to approve the pr to be able to merge it

PR #358, opened 2018-01-17
Cloud init improvements

indygreg:
Basket of changes. Read the commit messages.


deploy/template/etc/default/docker
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash -vex
 
+# Wait for cloud-init to finish initializing.
+if [ ! -e /var/lib/cloud/instance/boot-finished ]; then
```
walac: Shouldn't this be a `while` loop?

indygreg: Yes. Derp.

PR #359, opened 2018-01-17
Provide a script to create github releases

walac:
The old release.sh script is now called deploy.sh. The new release.sh is
intended to create a Github release of docker-worker.

By running it, it creates a tag which name is based on current date
and time, and uploads the worker-types-backup.json and
docker-worker-amis.jsonjson files.

release.sh is intended to run right after deploy.sh. The reason to put
them in two scripts is because we may want to test the new rolled out
AMIs first before doing a release, or we are just rolling out AMIs on
testing worker-types during development.

Notice the release script requires an environment variable called
DOCKER_WORKER_GITHUB_TOKEN with a user personal token set.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,29 +1,30 @@
 #!/bin/bash -e
 
-deploy_json=deploy/deploy.json;
+echo "Releasing docker-worker on Github"
 
-NODE_VERSION_MAJOR=$(node --version | tr -d v | awk -F. '{print $1}')
-NODE_VERSION_MINOR=$(node --version | tr -d v | awk -F. '{print $2}')
-
-if [ "$TASKCLUSTER_CLIENT_ID" == "" -o "$TASKCLUSTER_ACCESS_TOKEN" == "" ]; then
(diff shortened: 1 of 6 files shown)


release.sh
```diff
@@ -1,29 +1,30 @@
...
 
-NODE_VERSION_MAJOR=$(node --version | tr -d v | awk -F. '{print $1}')
-NODE_VERSION_MINOR=$(node --version | tr -d v | awk -F. '{print $2}')
-
-if [ "$TASKCLUSTER_CLIENT_ID" == "" -o "$TASKCLUSTER_ACCESS_TOKEN" == "" ]; then
-    echo "You don't seem to have proper Taskcluster credentials, please run 'taskcluster-cli signin' command" >&2
-    exit 1
+if [ "$DOCKER_WORKER_GITHUB_TOKEN" == "" ]; then
+  echo "You need to define the environment variable DOCKER_WORKER_GITHUB_TOKEN" >&2
+  echo "with a valida Github personal token." >&2
```
djmitche: *valid

walac:
> Looks good. Those worker-type-backup.json files don't contain worker-type secrets, right?

No, they don't.

PR #360, opened 2018-01-17
Knockout paravirtual AMIs

walac:
paravirtual AMIs are only used by the gecko-t-linux-medium worker type,
which receives no job.


indygreg:
This looks good to me!

PR #361, opened 2018-01-21
Bug 1431950 - check for onExitStatus before purgeCaches check

imbstack:
I believe this should fix the current issues with failing to report failed tasks.

### DIFF
--- a/src/lib/task.js
+++ b/src/lib/task.js
@@ -589,7 +589,7 @@ class Task extends EventEmitter {
         reporter = queue.reportFailed;
         taskState = 'failed';
       }
-      let purgeStatuses = this.task.payload.onExistStatus.purgeCaches;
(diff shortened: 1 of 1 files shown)


src/lib/task.js
```diff
@@ -589,7 +589,7 @@ class Task extends EventEmitter {
         reporter = queue.reportFailed;
         taskState = 'failed';
       }
-      let purgeStatuses = this.task.payload.onExistStatus.purgeCaches;
+      let purgeStatuses = this.task.payload.onExitStatus && this.task.payload.onExitStatus.purgeCaches;
```
jonasfj: I suggest doing: `(this.task.payload.onExitStatus || {}).purgeCaches || []` this way if `purgeCaches` is null the next thing still works... 

oh, nvm, I see you have a check there already...

PR #362, opened 2018-01-22
Increase the billing interval to 15 minutes

walac:


indygreg:
Do we actually have evidence for this? The shutdown timer doesn't start until the JS daemon starts. And once the JS daemon is running, it can start running tasks, right?

Where exactly is this "warm up time" race condition?

That being said, I'm fine increasing the billing cycle interval as a stop-gap so we get the new code deployed. But increasing to 1 hour is not proper. With the old code, we'd wait at most 1 hour after a task being terminating. This code would *always* wait 1 hour. So a 1 hour cooldown is on average going to keep workers around for longer. This is the wrong direction to move and defeats the purpose of the billing logic change. I think the cooldown should be shorter. Say 15 or 30 minutes (if you want to be conservative).

FWIW, I really think we want per-worker cooldowns. We want important workers like those that execute decision tasks to linger around for a while. We want expensive workers to terminate sooner so they don't run up the bill. We probably can keep cheaper workers around longer because they have a negligible impact on cost. For now, let's a pick a value that works for everyone and tweak later.

walac:
Ok, changed to 15 minutes.

PR #363, opened 2018-01-23
Bug 1427096 - Use mkdtemp to create the temporary folder for dind

glandium:
From irc:

> \<tomprince> glandium: I bet all you need to do is fix https://github.com/taskcluster/docker-worker/blob/master/src/lib/features/dind.js#L24 and https://github.com/taskcluster/docker-worker/blob/master/src/lib/features/dind.js#L39 to use a real temp file algo, rather than just picking a random directory and hoping it doesn't exist.'

Hopefully, this is what this does, but I don't have much clue in nodejs or docker-worker code, and I also don't know how to validate this actually works.

This bug has been a pain in my lower back for too long.

https://bugzilla.mozilla.org/show_bug.cgi?id=1427096

### DIFF
--- a/src/lib/features/dind.js
+++ b/src/lib/features/dind.js
@@ -21,7 +21,7 @@ class DockerInDocker {
     this.featureName = 'dind';
     // dind-service container
     this.container = null;
-    this.tmpFolder = path.join('/tmp', slugid.v4());
+    this.tmpFolders = [];
   }
 
   async link(task) {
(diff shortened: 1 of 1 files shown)


glandium:
More developments from irc:

> <tomprince> Hmm. https://github.com/taskcluster/docker-worker/blob/befe4fbc51e9ac7aec5e31f774facfe882517c7c/src/lib/features/dind.js#L11 leads me to https://github.com/isaacs/rimraf which suggests that `rmrf` is actually async. 
> <tomprince> Which, if link  -> kiilled -> link were all called in quick succession on the same Dind instance would cause the failures you've been seeing.

glandium:
Let's close this, hoping that #364 fixes it. Will reopen if it doesn't.

PR #364, opened 2018-01-23
Cleanups related to rimraf and mkdirp

glandium:
Cf. https://github.com/taskcluster/docker-worker/pull/363#issuecomment-359734466 this might actually fix bug 1427096 too, obsoleting PR #363.

### DIFF
--- a/src/lib/features/dind.js
+++ b/src/lib/features/dind.js
@@ -8,9 +8,8 @@
 
 const waitForSocket = require('../wait_for_socket');
 const slugid = require('slugid');
-const rmrf = require('rimraf');
 const path = require('path');
-const fs = require('fs');
+const {makeDir, removeDir} = require('../util/fs');
 let debug = require('debug')('docker-worker:features:dind');
 
 // Maximum time to wait for dind-service to be ready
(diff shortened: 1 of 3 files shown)


src/lib/volume_cache.js
```diff
@@ -2,32 +2,13 @@ var debug = require('debug')('docker-worker:volumeCacheManager');
 var path = require('path');
 var Promise = require('promise');
 var fs = require('fs');
-var mkdirp = require('mkdirp');
-var rmrf = require('rimraf');
+const {makeDir, removeDir} = require('util/fs');
```
walac: iirc, the correct syntax for importing the local module is `require('./util/fs')`

src/lib/features/dind.js
```diff
@@ -35,11 +34,7 @@ class DockerInDocker {
     debug('image verified %s', imageId);
 
     // Create temporary directory
-    await new Promise((accept, reject) => {
-      fs.mkdir(this.tmpFolder, (err) => {
-        return err ? reject(err) : accept();
-      });
-    });
+    await makeDir(self.tmpFolder);
```
walac: \<google\>Did you mean `this.tmpFolder`?\</google\>

glandium:
There are still failures, and in the case of the volume_cache ones, I really don't get why they happen as the changes are replacing functions with references to the exact same functions in another module.

walac:
@glandium no worries, I am digging into it.

PR #365, opened 2018-01-26
Fix volume cache purge based on exit status

walac:


src/lib/task.js
```diff
@@ -591,8 +591,8 @@ class Task extends EventEmitter {
       }
       let purgeStatuses = this.task.payload.onExitStatus && this.task.payload.onExitStatus.purgeCaches;
       if (purgeStatuses && purgeStatuses.includes(this.exitCode)) {
-        for (let cacheKey in this.task.volumeCaches) {
```
djmitche: I love this language.. [for-of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is for iterables, and `volumeCaches` is an array:

https://github.com/taskcluster/docker-worker/blob/befe4fbc51e9ac7aec5e31f774facfe882517c7c/src/lib/task.js#L106-L111

So this is a good fix.

PR #352, opened 2018-01-11
Change shutdown semantics based on per second billing

indygreg:
See the commit messages for more info.

PR #367, opened 2018-01-29
Fix AWS config test

walac:
PR #352 added a new configuration field, which, as a side effect, broke
AWS config test.


petemoore:
@walac Seems like a reasonable change to me.

PR #368, opened 2018-01-29
Upgrade docker

indygreg:
Let's run the latest/greatest version of Docker so we're more modern.


walac:
Will merge it after docker-worker next deployment

PR #369, opened 2018-01-30
Fix docker group for base AMI

walac:
PR #366 fixed docker group addition for Vagrant but broke the script for
AMI generation. This commit fixes it for both cases.

When building the AMI, we get the current user from $USER environment
variable. For Vagrant provisioning, we use $SUDO_USER.

### DIFF
--- a/deploy/packer/base/scripts/packages.sh
+++ b/deploy/packer/base/scripts/packages.sh
@@ -10,7 +10,17 @@ lsb_release -a
 
 # add docker group and add current user to it
 sudo groupadd -f docker
-sudo usermod -a -G docker $SUDO_USER
+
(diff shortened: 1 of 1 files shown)


deploy/packer/base/scripts/packages.sh
```diff
@@ -10,7 +10,18 @@ lsb_release -a
 
 # add docker group and add current user to it
 sudo groupadd -f docker
-sudo usermod -a -G docker $SUDO_USER
+
+env
```
djmitche: debugging?

walac: Ooops, fixed.

PR #374, opened 2018-02-08
Retry yarn install if it fails

walac:
For reasons I do not quite understand, buffertools sometimes fails to
build for no clear reason. As I am not willing to investigate the build
system of an external package, if "yarn install" fails, we just remove
the node_modules directory and try again.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -61,7 +61,7 @@ tasks:
       maxRunTime: 10800
       command:
         - >-
-          yarn install --frozen-lockfile && node_modules/.bin/eslint src test deploy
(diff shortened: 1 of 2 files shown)


.taskcluster.yml
```diff
@@ -61,7 +61,7 @@ tasks:
       maxRunTime: 10800
       command:
         - >-
-          yarn install --frozen-lockfile && node_modules/.bin/eslint src test deploy
+          while ! yarn install --frozen-lockfile; do rm -rf node_module; done && node_modules/.bin/eslint src test deploy
```
djmitche: missed one :)

PR #379, opened 2018-02-20
Bug 1427938: Mitigate Meltdown/Spectre

walac:
Canonical isn't going to patch kernel 4.4.0 on trusty for
Meltdown/Spectre bugs. This forces us to switch to linux-aws kernel, but
this kernel doesn't include the V4L2 and snd-aloop modules, necessary to
run tests Firefox on docker containers.

We now have our own custom kernel [1] and PPA [2]. This kernel is based
on the linux-aws, but with V4L2 and snd-aloop enabled.

[1] https://github.com/taskcluster/linux-aws-trusty
[2] https://launchpad.net/~taskcluster/+archive/ubuntu/ppa

### DIFF
--- a/deploy/packer/base/scripts/packages.sh
+++ b/deploy/packer/base/scripts/packages.sh
@@ -3,7 +3,7 @@
 set -e -v
 
 DOCKER_VERSION=17.12.0~ce-0~ubuntu
-KERNEL_VER=4.4.0-109-generic
+KERNEL_VER=4.4.0-1014-aws
 V4L2LOOPBACK_VERSION=0.10.0
 
 lsb_release -a
@@ -22,11 +22,12 @@ fi
 
 sudo usermod -a -G docker $user
 
-sudo apt-get update -y
-
 [ -e /usr/lib/apt/methods/https ] || {
   apt-get install apt-transport-https
 }
+
+sudo apt-get install -y software-properties-common
+sudo apt-add-repository ppa:taskcluster/ppa
 
 # Add docker gpg key and update sources
 sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
@@ -40,7 +41,6 @@ sudo apt-get update -y
 sudo apt-get install -y \
     linux-image-$KERNEL_VER \
     linux-headers-$KERNEL_VER \
-    linux-image-extra-$KERNEL_VER \
     dkms
 
 # Clean up old 3.13 kernel.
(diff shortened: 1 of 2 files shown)


indygreg:
Is this not a dupe of #375, which also switches to the AWS optimized kernel?

packages.sh already builds the V4L2 and snd-aloop modules. So it isn't necessary to build these modules into the system kernel package.

I'm very hesitant to run our own kernel. It's another piece of infrastructure that we need to maintain. We're going to have a pain keeping it in sync with the various upstream kernel improvements, including future security fixes. I don't think maintaining a kernel package is something we should be doing.

If the spectre/meltdown patches won't be applied to the official aws kernel image, then we're in a bind. If we need to use a custom kernel, we use a custom kernel. But I would strongly prefer using a pre-existing custom kernel instead of building our own. If we continue to build the v4l2 and snd-aloop modules as part of packages.sh, that means the only kernel variation is spectre/meltdown. I'm pretty sure we can find a PPA or something maintaining those patches.

FWIW, I have patches in progress to move to Ubuntu 16.04. I'll hopefully finish that in the next 2 weeks. Can we wait on that as our spectre/meltdown mitigation?

walac:
We don't build `V4L2` and `snd-aloop` modules, we only build `v4l2loopback` which depends on `V4L2`. `V4L2` and `snd-aloop` are built in tree. Meltdown and Spectre patches are applied to linux-aws, but `V4L2` and `snd-aloop` don't ship with this kernel.

I already tried upstream linux-aws and it didn't work because of the aforementioned issues. And I think it won't work on 16.04 either (although we can use the default kernel in this case).

indygreg:
I would rather we build the `V4L2` and `snd-aloop` as standalone modules (in `packages.sh`) rather than maintain our own kernel. Again, I don't think we should be in the business of maintaining our own kernel.

Unless these modules have to be built into the kernel, let me refactor #375 so these modules are built against the stock AWS kernel.

indygreg:
Also, whatever we end up doing, we should switch to the AWS optimized kernel permanently, as that should yield perf wins.

indygreg:
Poking at this some more, a PPA is probably sufficient for the time being. I was going to hack building the custom kernel [modules] into `packages.sh`. I did get that working. But it adds a lot of overhead. It is easier to manage the binary kernel package out of band.

I'll review this commit.

deploy/packer/base/scripts/packages.sh
```diff
@@ -3,7 +3,7 @@
 set -e -v
 
 DOCKER_VERSION=17.12.0~ce-0~ubuntu
-KERNEL_VER=4.4.0-109-generic
+KERNEL_VER=4.4.0-1014-aws
```
indygreg: Shouldn't this have `taskcluster` in its version to match the custom kernel version?

deploy/packer/base/scripts/packages.sh
```diff
@@ -127,7 +127,14 @@ sudo sh -c 'echo "options v4l2loopback devices=100" > /etc/modprobe.d/v4l2loopba
...
 echo "snd-aloop" | sudo tee --append /etc/modules
-sudo sh -c 'echo "options snd-aloop enable=1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1 index=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29" > /etc/modprobe.d/snd-aloop.conf'
+sudo sh -c 'echo "options snd-aloop enable=1,1,1,1,1,1,1,1 index=0,1,2,3,4,5,6,7" > /etc/modprobe.d/snd-aloop.conf'
+
+sudo bash -c 'cat > /etc/rc.local <<EOF
+#!/bin/sh -e
+modprobe snd-aloop
+exit 0
+EOF'
+sudo chmod +x /etc/rc.local
```
indygreg: This should be done in an `/etc/modprobe.d/` file.

deploy/packer/base/scripts/packages.sh
walac: I don't fully understand the apt matching algorithm, but the short answer is *no*. `apt-get` will manage to match the correct version, I checked that.

deploy/packer/base/scripts/packages.sh
walac: Err, forgot to add a comment here. I don't know why but in this new kernel putting `snd-aloop` inside `/etc/modules` doesn't load it at boot time, with no trace in `dmesg`. I put the `modprobe` in this file to make sure the module is correctly loaded.

PR #380, opened 2018-02-20
Generate apparmor profile in memory

walac:
As we upgraded docker, the default profile isn't stored in
/etc/apparmor.d/docker anymore. We now store the apparmor template for
docker as a constant variable, making agnostic regarding the docker
version.


walac:
Well, the reason I implemented like this is to make docker-worker less dependent on the external environment where it runs. And yes, the template was copied from the previous version.

PR #381, opened 2018-02-26
Bug 1339236 - use task.scopes instead of cert.scopes for proxy

djmitche:


src/lib/features/taskcluster_proxy.js
```diff
@@ -37,9 +37,9 @@ class TaskclusterProxy {
     if (task.claim.credentials.certificate) {
       cmd.push('--certificate=' + task.claim.credentials.certificate);
     }
-    const cert = JSON.parse(task.claim.credentials.certificate);
-    cmd.push(`--task-id=${task.status.taskId}`);
```
walac: Is the `--task-id` parameter not needed anymore or this was a typo?

src/lib/features/taskcluster_proxy.js
```diff
@@ -37,9 +37,9 @@ class TaskclusterProxy {
...
-    cmd = cmd.concat(cert.scopes);
+
+    // supply the tsak's scopes, limiting what can be done via the proxy
```
walac: nit: `task's`

src/lib/features/taskcluster_proxy.js
djmitche: Correct, it's not needed since we are providing the information the proxy might need from the task (its scopes).

src/lib/features/taskcluster_proxy.js
djmitche: fixed

djmitche:
I don't know of a good way to test this.  The concern is that there might be tasks out there using the additional scopes returned by `claimWork` beyond `task.scopes` -- specifically, scopes to upload artifacts to the running task.  I don't *think* this is the case, but I can't see a way to find out..

walac:
> I don't know of a good way to test this. The concern is that there might be tasks out there using the > additional scopes returned by claimWork beyond task.scopes -- specifically, scopes to upload artifacts to the running task. I don't think this is the case, but I can't see a way to find out..

What I do is deploy in ami-test and run a try push to it. If everything is ok, roll out to production :)

djmitche:
Running in try should be a good test.

The test failure here is
```
  1) artifact extraction tests upload retry:
```
so I think that's unrelated to my work?

walac:
> so I think that's unrelated to my work?

No, it still fails sometimes.

djmitche:
@walac was there something preventing merging this?

walac:
Merged :)

PR #383, opened 2018-03-05
Bug 1302596: Use promise-retry for artifact download

walac:


src/lib/util/artifact_download.js
```diff
@@ -29,78 +31,58 @@ module.exports = async function(queue, stream, taskId, artifactPath, destination
...
+    blank: ' ',
+  });
 
-      if (req.response.statusCode !== 200) {
-        let error = new Error(req.response.statusMessage);
-        error.statusCode = req.response.statusCode;
-        throw error;
-      }
+  return await promiseRetry((retry, attempt) => {
+    return new Promise(async(accept, reject) => {
```
djmitche: I think the Promise constructor ignores the return value of its argument.  In this case, that's another Promise (because it's an async function).  If anything in the body throws an exception, it will be converted to a rejection of that second Promise, which will go unhandled.

I think this could be

```javascript
return await promiseRetry(async (retry, attempt) => {
  try {
    let hash = crypto.createHash('sha256');
    // ..
    return `sha256:${hash.digest('hex')}`; 
  } catch (err) {
    debug(`Error downloading "${artifactPath}" from task ID "${taskId}". ${e}`);
    // ...
    retry(err);
  }
}, {
   retries: maxAttempts,
   factor: delayFactor,
   randomize: randomizationFactor,
});
```

src/lib/util/artifact_download.js
```diff
@@ -29,78 +31,58 @@ module.exports = async function(queue, stream, taskId, artifactPath, destination
...
+      let destinationStream = fs.createWriteStream(destination);
 
-      if (receivedSize !== expectedSize) {
-        throw new Error(`Expected size is '${expectedSize}' but received '${receivedSize}'`);
-      }
+      let sink = got.stream(artifactUrl)
+        .on('data', chunk => hash.update(chunk))
+        .on('error', reject)
+        .on('downloadProgress', progress => bar.update(progress.percent));
+      await pipe(sink, destinationStream);
```
djmitche: My understanding of streams is terrible, but .. don't you need to wait for this stream to complete?  I don't think `pipe` actually returns anything, so awaiting it doesn't do so.  But maybe I'm wrong?

walac: This uses the [promisepipe](https://github.com/epeli/node-promisepipe) package, the `await` keyword handles it.

djmitche: Ah, indeed, I am wrong :)

PR #388, opened 2018-04-09
bug 1452266 - indent chainOfTrust.json

escapewindow:
gpg signature verification has a max line length of 20k chars. By using newlines and indents, we avoid hitting this limit.

### DIFF
--- a/src/lib/features/chain_of_trust.js
+++ b/src/lib/features/chain_of_trust.js
@@ -95,7 +95,7 @@ class ChainOfTrust {
     });
 
     let signedChainOfTrust = await openpgp.sign({
-      data: JSON.stringify(certificate),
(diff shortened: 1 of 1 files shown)


escapewindow:
I'm not sure if this is correct. I'm mainly using https://stackoverflow.com/questions/5670752/how-can-i-pretty-print-json-using-node-js .

djmitche:
@walac I merged this but did not make a release.

walac:
Is this urgent somehow?

djmitche:
We thought so, but I think @escapewindow found another way..

escapewindow:
We found another way around for mozilla-beta Fx60, but our partner repack migration work on maple is blocked on this. Partner repacks are the final piece left on buildbot other than esr52.

PR #390, opened 2018-04-19
Bug 1455117 - check supersederUrl before fetching it

djmitche:


walac:
It feels like it broke a bunch of tests. Chunk 1 seems suspicious, I retriggered it.

djmitche:
Looks like it was a regex-quoting issue.  I've pushed a new change.

djmitche:
Third time is the charm?

djmitche:
Yup, that single error looks unrelated.

PR #393, opened 2018-05-25
Bug 1156779: Fail the task if the payload contains an invalid feature

walac:
The current behavior is to run the task even if an invalid feature is
passed in the payload. We change this to fail the task if an invalid
feature is detected.

### DIFF
--- a/src/lib/task.js
+++ b/src/lib/task.js
@@ -114,6 +114,13 @@ function buildStateHandlers(task, monitor) {
   let handlers = [];
   let featureFlags = task.payload.features || {};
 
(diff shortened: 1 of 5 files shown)


src/lib/task.js
```diff
@@ -114,6 +114,13 @@ function buildStateHandlers(task, monitor) {
   let handlers = [];
   let featureFlags = task.payload.features || {};
 
+  // (featureFlags - features) gives the set of non supported requested features
+  let diff = _.filter(_.map(featureFlags, (x, i) => i), x => !features.hasOwnProperty(x));
```
djmitche: The `_.map` here would be better written `_.keys(featureFlags)`

PR #397, opened 2018-05-31
Bug 1458737: Reboot the instance on OOM

walac:
When OOM Killer is triggered, it can kill any process in the system. The
safest measure it to reboot the instance.


indygreg:
Is it possible to shut down the instance on OOM? Reboot on OOM is better than current behavior. I only see risks involved with rebooting though. What if the shutdown wasn't clean and the host reboots in a bad state?

djmitche:
Not that i could find in my digging around in kernel config :(

walac:
The reboot will trigger a fail in getting secrets, which will then trigger an instance termination

PR #404, opened 2018-06-20
partial update of README with more details of deploying

djmitche:
So this helped me to create `deploy.json`, but after that packer got "stuck".  90 minutes later, I finally decided to use 'strace' to see what it was up to, and it was stuck at

```
dustin@jemison ~ $ strace -p 12474
strace: Process 12474 attached
read(0, ^Cstrace: Process 12474 detached
 <detached ...>
```

so I tried typing things:
```
packer build -var-file deploy/target/packer/app.json -var templateContents=deploy/deploy.tar.gz -var npmPackage=docker-worker.tgz -var workerRevision=ef974e09acbf6d36c641af0468893e9355f951cc deploy/packer/app.json

skipping line: 1

skipping line: 2

skipping line: 3

skipping line: 4

skipping line: 5

skipping line: 6
please build me a docker-worker?
pweeeeeze?

skipping line: 9

skipping line: 10
so, no?
11 3

real    93m11.387s
user    0m0.002s
sys     0m0.000s
```
after "so, no?" I hit ctrl-D, at which time it printed "11 3" and the time output.

So, I have no idea what's going on here, but the build process didn't build.  So these instructions may still be incomplete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ are for hacking on the worker itself.
   - Node >= 8.5.0
   - Docker
   - Packer (to build AMI)
+    - Note that cracklib also hsa a utility named `packer` - run `packer version` to check!
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


djmitche:
That issue was because I was running 'packer' from cracklib, rather than Hashicorp's packer. OOPS.

djmitche:
@walac OK, pushed some README and script changes.  What do you think?

PR #408, opened 2018-07-25
Bug 1478394 - upgrade docker

djmitche:
I commented out the `rm /var/lib/docker` because something was holding it open.  That likely leaves a bunch of data in the AMI, but it's not the end of the world.


djmitche:
Since I already (accidentally) released this..

PR #411, opened 2018-09-14
Bug 1488148: Add localhost.localdomain host name

walac:
This is a requirement for nss-try. We could work on allowing additional
configuration from the VM using the user data, but as we aim to replace
docker-worker, lets postpone more elaborate configs for newer worker.

### DIFF
--- a/schemas/payload.json
+++ b/schemas/payload.json
@@ -38,6 +38,11 @@
   ],
   "additionalProperties": false,
   "properties": {
+    "log": {
+      "title": "Custom log location",
+      "description": "Specifies a custom location for the livelog artifact",
+      "type": "string"
+    },
     "image": {
       "title": "Docker image.",
       "description": "Image to use for the task.  Images can be specified as an image tag as used by a docker registry, or as an object declaring type and name/namespace",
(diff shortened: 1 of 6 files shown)


schemas/payload.json
```diff
@@ -38,6 +38,10 @@
   ],
   "additionalProperties": false,
   "properties": {
+    "log": {
+      "title": "Custom log location",
+      "type": "string"
```
djmitche: I don't see any changes to actually use this value -- was this diff included by mistake?

If not, a `description` property would be great: What does this mean?  Is it a path on the host or in the docker container?  Or is it the artifact name?

src/lib/task.js
```diff
@@ -400,7 +400,10 @@ class Task extends EventEmitter {
         Env: taskEnvToDockerEnv(env),
         HostConfig: {
           Privileged: privilegedTask,
-          ShmSize: 1800000000
+          ShmSize: 1800000000,
+          ExtraHosts: [
+            'localhost.localdomain:127.0.0.1', // for nss-try
+          ],
```
djmitche: This seems like the critical change, and seems fine (maybe link to the bug instead of just saying it's for nss-try? It seems pretty generic and other projects might want the functionality too)

schemas/payload.json
walac: This is part of unit tests fixes broken in #409

djmitche: Ah, ok -- could you add the description then?  I didn't know such a thing was available (I still don't!)

PR #416, opened 2018-10-29
Bug 1501720: Use v5.0 dind-service

walac:
v5.0 uses the same docker version as docker-worker.


djmitche:
Sorry for the delay -- I got a lot of review requests!

PR #417, opened 2018-10-30
Bug 1496906: Add more logs details to artifact upload

walac:


src/lib/upload_to_s3.js
```diff
@@ -112,7 +114,7 @@ module.exports = async function uploadToS3 (
         });
 
         req.on('error', err => {
-          log(`Error uploading ${artifactName}`, logDetails);
+          log(`Error uploading ${artifactName}`, _.defaults({err}, logDetails));
```
edmorley: The docker-worker readme says the minimum Node.js it supports is 8.5.0, which means that it should be possible to use [spread properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) (they were supported [as of 8.3.0](https://node.green/#ES2018-features-object-rest-spread-properties-object-spread-properties)), avoiding the need for lodash?

```js
log(`Error uploading ${artifactName}`, {err, ...logDetails});
```

walac: Good point

PR #418, opened 2018-11-19
Bug 1508383 - pass TASKCLUSTER_{ROOT,PROXY}_URL to tasks

djmitche:
Per https://github.com/taskcluster/taskcluster-rfcs/pull/128

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -53,7 +53,7 @@ defaults:
   restrictCPU: false
 
   # Image used to  create the taskcluster proxy container.
-  taskclusterProxyImage: 'taskcluster/taskcluster-proxy:4.0.1'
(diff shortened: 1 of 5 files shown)


walac:
It feels like there is a regression

djmitche:
Looking now..

djmitche:
Ah, I think it's because https://github.com/taskcluster/taskcluster-proxy/pull/38 hasn't landed yet, so when it starts taskcluster-proxy, the proxy bails because it doesn't understand `--root-url`.  Let's wait until that PR lands.

djmitche:
OK, upgraded to the just-released 5.0.0 from https://github.com/taskcluster/taskcluster-proxy/pull/38.  Let's see what the tests do..

PR #420, opened 2018-12-13
Add a handler for the error event on the readStream, too

djmitche:


djmitche:
@walac mentioned this should be using streampipe instead.  I'll give that a shot.

PR #421, opened 2018-12-27
Use the correct userdata property for rootUrl

djmitche:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1462768 this property
is named `taskclusterRootUrl`.  And since that's landed, we can remove the
default.

### DIFF
--- a/src/lib/host/aws.js
+++ b/src/lib/host/aws.js
@@ -115,6 +115,11 @@ module.exports = {
     let userdata = await getJsonData(`${baseUrl}/user-data`);
     let securityToken = userdata.securityToken;
     let rootUrl = userdata.taskclusterRootUrl;
+    if (!rootUrl) {
+      // Without a rootUrl, there's no way to claim tasks, so bail out..
+      log('[alert-operator] no rootUrl provided');
+      spawn('shutdown', ['-h', 'now']);
+    }
 
     log('read userdata', { text: userdata });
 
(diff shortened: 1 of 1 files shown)


djmitche:
https://tools.taskcluster.net/groups/PEvGrCbUSh24hm45v6008g/tasks/PEvGrCbUSh24hm45v6008g/runs/0/logs/public%2Flogs%2Flive.log confirms that this is, indeed, the property name.

src/lib/host/aws.js
```diff
@@ -114,12 +114,7 @@ module.exports = {
 
     let userdata = await getJsonData(`${baseUrl}/user-data`);
     let securityToken = userdata.securityToken;
-
-    let rootUrl = userdata.rootUrl;
-    if (!rootUrl) {
-      // for backward compatibility, assume the legacy instance if no rootUrl is given
-      rootUrl = 'https://taskcluster.net';
-    }
+    let rootUrl = userdata.taskclusterRootUrl;
```
walac: What if `taskclusterRootUrl` isn't defined?

djmitche: Is an assert the best way to bail out at this point?

djmitche: Ah, I see a call to `shutdown -h now` a few lines later..

djmitche:
The one task that failed in that CI run was for a 500 ISE from Docker.  I re-ran it.

djmitche:
green now :)

src/lib/host/aws.js
djmitche: In a few months, that won't work, though.  Does it make sense at that point to have a default value that doesn't work?

walac: Fair enough :)

PR #425, opened 2019-01-18
Bug 1509311: Initial work on packet.net custom image

walac:


deploy/template/lib/systemd/system/docker-worker.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Taskcluster docker worker
+Requires=docker.service
+
+[Service]
+Type=notify
+ExecStart=/usr/local/bin/node /root/docker_worker/src/bin/worker.js --host test production
```
djmitche: will this be `--host packet` later?

deploy/bin/build
```diff
@@ -46,8 +46,16 @@ cmd_app() {
     --exclude='./worker-types-backup.json' \
     --exclude='./docker-worker-amis.json' \
     --exclude='./packer-artifacts.json' \
+    --exclude='docker-worker.tar.gz' \
+    --exclude='modules.tar.gz' \
+    --exclude='kernel.tar.gz' \
+    --exclude='initrd.tar.gz' \
```
djmitche: Would it be better to whitelist the subdirectories required here?  I've been bitten multiple times by things that tar up a repo when I have some temporary file there containing secrets..

deploy/template/lib/systemd/system/docker-worker.service
walac: Yep, I need to work on the packet-net branch

deploy/bin/build
walac: Yep, I can do that in a follow up PR

PR #426, opened 2019-01-22
[packet] Do not add tar files to the image

walac:
docker-docker.tgz and image.tar.gz were mistakenly added to the package.

### DIFF
--- a/deploy/bin/build
+++ b/deploy/bin/build
@@ -46,7 +46,8 @@ tar_docker_worker() {
     --exclude='./worker-types-backup.json' \
     --exclude='./docker-worker-amis.json' \
     --exclude='./packer-artifacts.json' \
-    --exclude='docker-worker.tar.gz' \
+    --exclude='docker-worker.tgz' \
+    --exclude='image.tar.gz' \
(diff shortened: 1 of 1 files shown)


deploy/bin/build
```diff
@@ -46,7 +46,8 @@ tar_docker_worker() {
     --exclude='./worker-types-backup.json' \
     --exclude='./docker-worker-amis.json' \
     --exclude='./packer-artifacts.json' \
-    --exclude='docker-worker.tar.gz' \
+    --exclude='docker-worker.tgz' \
+    --exclude='image.tar.gz' \
```
djmitche: Might it make more sense to turn this into a whitelist?  I always get scared that blacklists like this are going to pick up a stray `root-password.txt` or something in my working copy..

djmitche: Where by "worry" I mean that I think I've had three security incidents along those lines, the most recent just two weeks ago..

walac: Yeah, I never understood why it wasn't a white list in the first place, once I have some time I will change this.

PR #429, opened 2019-01-26
bug 1518912 - add ed25519 cot signature support

escapewindow:
This should roll out together with https://github.com/taskcluster/generic-worker/pull/137 and https://github.com/mozilla-releng/scriptworker/pull/294 .


escapewindow:
Trying to test. We still require vagrant? https://github.com/taskcluster/docker-worker#setting-up-vagrant-recommended

escapewindow:
It looks like we need to update the `yarn.lock` file. When I ran `yarn install` without `--frozen-lockfile` since it breaks on the missing `tweetnacl`. This works, but blows away all the `integrity` lines.

escapewindow:
I had to set the `TASKCLUSTER_ROOT_URL` env var manually inside vagrant; also, the tc-signin env vars didn't populate, so I copy-pasted from another tmux pane. Looks like I've got tests running now!

escapewindow:
> It looks like we need to update the `yarn.lock` file. When I ran `yarn install` without `--frozen-lockfile` since it breaks on the missing `tweetnacl`. This works, but blows away all the `integrity` lines.

Looks like I need [`yarn>=1.10.0`](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/22027)

walac:
I am assigning to myself to not forget about it. @escapewindow give me a heads up on IRC when it is ready for review, pls :)

escapewindow:
@walac, I'm having a hard time with the integration test.

I'm able to verify the signature using `test/fixtures/ed25519_public_key` in python, so `public/chain-of-trust.json.sig` is a valid signature for `public/chain-of-trust.json`, and the json+sig artifact upload process works for both wget and Firefox to download valid artifacts.

However, when I download `public/chain-of-trust.json.sig` in the integration test via `getArtifact`, I get 60 bytes instead of 64. I've tried various things to try to debug: gunzip, `console.log` a base64 version of the sig so I can try to see what is wrong with it, switch to the `request` and `https` modules for the sig, etc.

Do you have any idea what might be going wrong for me?

walac:
Let me check this

escapewindow:
OMG, thank you!
I think I'm ready for review now.

escapewindow:
> @escapewindow the problem was with content encoding, it was downloading as a text, so any `null` terminator could end the stream sooner. This is a patch that makes it works, I am using the [got](https://github.com/sindresorhus/got) library because it makes things easier.

Does that mean my upload has problems, or just that the integration `getArtifact` call assumes text?

deploy/bin/gen-ed25519-key.js
```diff
@@ -0,0 +1,16 @@
...
+function main() {
+  console.log('Generating ed25519 keypair...');
+  let keypair = tweetnacl.sign.keyPair();
+
+  fs.writeFile('docker-worker-ed25519.pub', new Buffer.from(keypair.publicKey).toString('base64'), function(err) {});
+  return fs.writeFile('docker-worker-ed25519-cot-signing-key.key', new Buffer.from(keypair.secretKey).toString('base64'), function(err) {});
+}
+
+// can't do this b/c `TypeError: Cannot read property 'catch' of undefined`
+// main().catch(err => console.log(err.stack));
```
walac: this is because `main` is not an `async` function, so it doesn't return a promise.

deploy/bin/gen-ed25519-key.js
```diff
@@ -0,0 +1,16 @@
...
+
+const fs = require('fs');
+const tweetnacl = require('tweetnacl');
+
+function main() {
+  console.log('Generating ed25519 keypair...');
+  let keypair = tweetnacl.sign.keyPair();
+
+  fs.writeFile('docker-worker-ed25519.pub', new Buffer.from(keypair.publicKey).toString('base64'), function(err) {});
+  return fs.writeFile('docker-worker-ed25519-cot-signing-key.key', new Buffer.from(keypair.secretKey).toString('base64'), function(err) {});
```
walac: Maybe it is better to use [writeFileSync](https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options) here.

deploy/packer/app.json
```diff
@@ -43,10 +43,20 @@
       "destination":    "docker-worker-gpg-signing-key.key",
       "only":           ["hvm-builder-trusted"]
     },
+    {
+      "type":           "file",
+      "source":         "{{user `cotEd25519SigningKey`}}",
+      "destination":    "docker-worker-ed25519-signing-key.key",
+      "only":           ["hvm-builder-trusted"]
+    },
```
walac: We need to create/copy a fake key for `hvm-builder`, don't we?

src/lib/features/chain_of_trust.js
```diff
@@ -25,6 +26,8 @@ class ChainOfTrust {
     this.hash = crypto.createHash('sha256');
     let armoredKey = fs.readFileSync(task.runtime.signingKeyLocation, 'ascii');
     this.key = openpgp.key.readArmored(armoredKey).keys;
+    let sk = Buffer.from(fs.readFileSync(task.runtime.ed25519SigningKeyLocation, 'ascii'), 'base64');
+    this.ed25519Key = Buffer.from(fs.readFileSync(task.runtime.ed25519SigningKeyLocation, 'ascii'), 'base64');
```
walac: Here I think it is better to use the `async` version of these function, to not harm concurrency. The best way is to use promises, like so:
```javascript
let sk = Buffer.from(await new Promise((accept, reject) => 
  fs.readFile('path', (err, data) => err ? reject(err) : accept(data))), 'base64');
```

src/lib/features/chain_of_trust.js
```diff
@@ -93,9 +96,15 @@ class ChainOfTrust {
         certificate.environment[tag] = task.runtime[tag];
       }
     });
+    let chainOfTrust = JSON.stringify(certificate, null, 2);
+    let chainOfTrustSig = tweetnacl.sign.detached(Buffer.from(chainOfTrust), this.ed25519Key);
+    var cotBufferStream = new stream.PassThrough();
```
walac: nit: s/`var`/`let`/

src/lib/features/chain_of_trust.js
```diff
@@ -93,9 +96,15 @@ class ChainOfTrust {
...
+    cotBufferStream.end(Buffer.from(chainOfTrust));
+    var sigBufferStream = new stream.PassThrough();
```
walac: nit: s/`var`/`let`/

walac: What's the purpose of `sigBufferStream`?

src/lib/features/chain_of_trust.js
walac: What's the purpose of `cotBufferStream`?

src/lib/features/chain_of_trust.js
```diff
@@ -113,6 +122,29 @@ class ChainOfTrust {
...
+    try {
+      await uploadToS3(task.queue, task.status.taskId, task.runId,
+        sigBufferStream, 'public/chain-of-trust.json.sig', expiration, {
+          'content-type': 'application/octet-stream',
+          'content-length': chainOfTrustSig.length
+        });
+    } catch (err) {
+      debug(err);
+      throw err;
+    }
```
walac: the `try`/`catch` pairs aren't necessary here, the errors will be reported to either Papertrail or the task log.

test/integration/chain_of_trust.js
```diff
@@ -61,13 +64,25 @@ suite('certificate of trust', () => {
...
     assert.deepEqual(Object.keys(result.artifacts).sort(), expectedArtifacts);
 
+    // ed25519 cot
+    let chainOfTrust = await getArtifact(result, 'public/chain-of-trust.json');
+    let queue = new taskcluster.Queue();
+    let url = queue.buildUrl(queue.getArtifact, result.taskId, result.runId, 'public/chain-of-trust.json.sig');
+    let chainOfTrustSig = (await got(url, {encoding: null})).body;
+
+    let verifyKey = Buffer.from(fs.readFileSync('test/fixtures/ed25519_public_key', 'ascii'), 'base64');
+    assert(tweetnacl.sign.detached.verify(new Uint8Array(Buffer.from(chainOfTrust)), Buffer.from(chainOfTrustSig), verifyKey), 'ed25519 chain of trust signature does not appear to be valid');
```
walac: iiuc the `tweetnacl` docs correctly, the `Uint8Array` isn't necessary, since for nodejs >= v4 we can use the `Buffer` types.

src/lib/features/chain_of_trust.js
walac: The `ascii` argument comes before the callback :)

escapewindow: Thanks! Fixed. I think I've addressed all of your concerns now.

deploy/bin/gen-ed25519-key.js
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const tweetnacl = require('tweetnacl');
+
+function main() {
+  console.log('Generating ed25519 keypair...');
+  let keypair = tweetnacl.sign.keyPair();
+
+  fs.writeFile('docker-worker-ed25519.pub', new Buffer.from(keypair.publicKey).toString('base64'), function(err) {});
```
walac: This also should be a sync version

escapewindow: Done.

escapewindow:
Looks like docker-worker timeouts after complaining about yarn.lock. Related? Or maybe related to the high queue load last week?

walac:
I think it is related to the yarn issue

escapewindow:
Hopefully that fixes it.

escapewindow:
@walac could you close and reopen? Hoping the tests green up, or we get to the next error(s).

PR #430, opened 2019-01-28
Bug 1522852: Implement packet.net cloud provider

walac:


src/lib/host/packet.js
```diff
@@ -0,0 +1,94 @@
...
+        .filter(addr => addr.address_family === 4 && addr.public)
+        .map(addr => addr.address)[0];
+      const privateIp = data.network.addresses
+        .filter(addr => addr.address_family === 4 && !addr.public)
+        .map(addr => addr.address)[0];
+
+      assert(publicIp);
+      assert(privateIp);
+
+      const userdata = fs.readFileSync('/var/lib/cloud/instance/user-data.txt')
```
djmitche: what creates this file?

src/lib/host/packet.js
```diff
@@ -0,0 +1,94 @@
...
+        clientId: userdata.TASKCLUSTER_CLIENT_ID,
+        accessToken: userdata.TASKCLUSTER_ACCESS_TOKEN,
+      };
+
+      const secrets = new taskcluster.Secrets({
+        rootUrl: userdata.rootUrl,
+        credentials,
+      });
+
+      const secretsData = await secrets.get('project/taskcluster/docker-worker:secrets');
```
djmitche: It might make sense to parameterize this secret name, or even just include the whole name in userdata.  Then different workerTypes could get different secrets, if necessary.

src/lib/host/packet.js
walac: packet deployment

PR #432, opened 2019-02-05
Bug 1439110: Omit accessToken from Papertrail logs

walac:
Papertrail logs are restrict to core members, but even so let's harden
it.

### DIFF
--- a/src/lib/features/taskcluster_proxy.js
+++ b/src/lib/features/taskcluster_proxy.js
@@ -105,7 +105,7 @@ class TaskclusterProxy {
             task.runtime.log('Credentials updated', {
               taskId: task.status.taskId,
               runId: task.runId,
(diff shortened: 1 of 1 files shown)


src/lib/features/taskcluster_proxy.js
```diff
@@ -105,7 +106,7 @@ class TaskclusterProxy {
             task.runtime.log('Credentials updated', {
               taskId: task.status.taskId,
               runId: task.runId,
-              credentials: creds
+              credentials: _.omit(creds, ['accessToken']),
```
djmitche: it'd probably be good to omit the certificate, too.  Maybe just `clientId: creds.clientId`?

walac: Done.

PR #438, opened 2019-02-28
Bug 1531320 - Update password store secrets (git pull) in deploy script

petemoore:
This is rather simple, it doesn't do a `git clean` or `git reset` in case there are local changes that people are working on, but it seems sane to do at least a `git pull` to make sure the secrets are the latest (as it is easy to forget to refresh them when deploying).

See [bug 1531320](https://bugzil.la/1531320) for details.


petemoore:
Closing. See bug link for details.

PR #439, opened 2019-03-14
Bug 1535179 - add two new workerTypes

djmitche:


config.yml
```diff
@@ -42,7 +42,7 @@ defaults:
 
   statelessHostname:
     enabled: false
-    secret: process.env.DNS_SERVER_SECRET
+    secret: !env DNS_SERVER_SECRET
```
petemoore: i'm not familiar with this syntax so getting a review from someone else for this line would be great

deploy/etc/worker-types.json
```diff
@@ -1,5 +1,7 @@
 {
   "hvm-builder": [
+    "android-components-g",
+    "application-services-r",
```
petemoore: Based on [this report](https://bugzilla.mozilla.org/show_bug.cgi?id=1535352#c0) I suspect we should add some more that just these, but this is a good start for [bug 1535179](https://bugzil.la/1535179), and we can probably add the others in [bug 1535352](https://bugzil.la/1535352).

djmitche: I think @tomprince is adding a bunch more.

config.yml
djmitche: To be honest, we could just drop this, as nothing ever provides the env var (or if it did, it didn't work).  This is at least less-awfully-broken than the previous :)

PR #441, opened 2019-03-28
bug 1534506 - Remove cot openpgp usage.

escapewindow:
I believe we still use openpgp for encrypted env vars, so I left that in place. I just removed any cot gpg references.

This can roll out at any time. If this lands before we rotate the gpg keypair, we can just get rid of the gpg keypair.


escapewindow:
@walac putting this on your radar. No rush, though the sooner we land this, the sooner we can ditch the cot openpgp keypair.

escapewindow:
> There are more places to change:
> 
>     * Remove the [script](https://github.com/taskcluster/docker-worker/blob/master/deploy/bin/gen-signing-key) to create the keys
> 
>     * Remove it from [deployment scripts](https://github.com/taskcluster/docker-worker/blob/master/deploy/packer/app.json#L60-L79)

Are we sure this isn't the encrypted env vars key?

> Also, you need to rebase your branch to get tests green.

Done.

escapewindow:
Removed the signingKeyLocation and gen-signing-key refs.

walac:
Some more places: [here](https://github.com/taskcluster/docker-worker/blob/master/deploy/bin/import-docker-worker-secrets#L52-L53), [here](https://github.com/taskcluster/docker-worker/blob/master/deploy/bin/import-docker-worker-secrets#L66), [here](https://github.com/taskcluster/docker-worker/blob/master/deploy/packer/app.json#L60-L65), [here](https://github.com/taskcluster/docker-worker/blob/master/deploy/packer/app.json#L72-L75), [here](https://github.com/taskcluster/docker-worker/blob/master/deploy/template/packer/app.json#L3) and [here](https://github.com/taskcluster/docker-worker/blob/master/deploy/variables.js#L51-L53).

I am quite sure I am missing a spot, but I handle it if there one.

escapewindow:
Done, and just now fixed the recent merge conflicts.

escapewindow:
Looks like we need a `let verified = Buffer.from(chainOfTrust);` or something? Not sure if we can reuse the buffer like that, or if we need to write it to disk or what.

walac:
> Looks like we need a let verified = Buffer.from(chainOfTrust); or something? Not sure if we can reuse the buffer like that, or if we need to write it to disk or what.

I think we can't, the `Buffer.from` value must be saved in a variable.

escapewindow:
This may have fixed it. My local tests in vagrant are timing out :\

escapewindow:
@walac do you know if you want these changes? If so, we can keep working on this PR. If not, we can leave this until generic-worker replaces docker-worker. (That means we'll still create gpg-signed cot artifacts in docker-worker, even though they're obsolete.)

walac:
@escapewindow Oh, sorry, I am so much with heads down with other stuff that I forgot your PR. Yes, definitely this is worth. I will re-review it soon. I am so sorry again.

escapewindow:
Thanks!

PR #443, opened 2019-04-10
Bug 1527107: Fix CVE-2019-5736

walac:
docker-ce >= 18.09.2 fixes this vulnerability. Unfortunatelly, docker
channel doesn't distribute this version for Ubuntu trusty. As trusty
support is expiring too, we update the Ubuntu version in the AMI to
bionic.

The default kernel for bionic in EC2 is the linux-image-aws, but it
doesn't come with V4L2, necessary for some tests. We then switch to the
generic kernel.

Notice bionic uses systemd, and we have to upgrade startup scripts to
cope with it. We had to modify some systemd scripts to break dependency
cycles.


deploy/packer/base/scripts/packages.sh
```diff
@@ -2,9 +2,9 @@
 
 set -e -v
 
-DOCKER_VERSION=18.06.0~ce~3-0~ubuntu
-KERNEL_VER=4.4.0-1014-aws
-V4L2LOOPBACK_VERSION=0.10.0
+DOCKER_VERSION=18.03.1~ce~3-0~ubuntu
```
djmitche: 18.03??

deploy/template/usr/local/bin/start-docker-worker
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+DOCKER_WORKER='node /home/ubuntu/docker_worker/src/bin/worker.js'
+DOCKER_WORKER_OPTS=
+META='http://170.254.169.254/latest/meta-data/'
```
djmitche: looks like this is unused?

deploy/packer/base/scripts/packages.sh
walac: It was a mistake, I am fixing it

deploy/template/usr/local/bin/start-docker-worker
walac: Hrm, just to make it equal to its `upstart` counterpart.

djmitche: Does it need `export` then?

walac: It the original it hadn't

djmitche:
I'm going to merge now to avoid conflicts, but wcosta will be deploying later.

PR #456, opened 2019-05-21
Bug 1552617 - use instance name for worker-id

grenade:
the instance worker id is being rounded at some point in its journey to the queue. if we use the instance name (which is prefixed with "vm-"), it should prevent integer rounding by virtue of being an obvious string.


walac:
Thanks!

PR #457, opened 2019-05-24
Fix packet.net image build

walac:


djmitche:
I think the last part was the important bit here, where it gets data from userdata?

walac:
> I think the last part was the important bit here, where it gets data from userdata?

You mean, who provides the userdata to the instance?

djmitche:
No, I was just indicating that the part I looked closely at was the last bit.  It looks like the Dockerfile changes were unrelated?  But it all seems fine to land.  Sorry for the confusion :)

PR #458, opened 2019-05-28
Move certificate to repo

walac:
The certificate is public available and there is no problem in storing
it in the repo.


walac:
> It doesn't look like this actually adds a certificate?

Oops, added, thanks!

PR #459, opened 2019-05-30
Bug 1545308: Run packet instances with performance CPU governor 

walac:
By default, the CPU governor is set to powersave, and the cores can run
in very low frequencies. Setting to performance makes the cores run at
their maximum frequencies.


walac:
> This looks good -- will that set-cpufreq script get run automatically? I don't see it referenced anywhere.

It is part of the system, I am replacing it adding the `performance` branch. It is called by systemd.

PR #464, opened 2019-06-07
Bug 1546686 - allow configuration by taskcluster-worker-runner

djmitche:
And remove some false info from usage

### DIFF
--- a/deploy/template/usr/local/bin/start-docker-worker
+++ b/deploy/template/usr/local/bin/start-docker-worker
@@ -1,6 +1,3 @@
 #!/bin/bash
 
-DOCKER_WORKER='node /home/ubuntu/docker_worker/src/bin/worker.js'
-DOCKER_WORKER_OPTS=
-
-$DOCKER_WORKER $DOCKER_WORKER_OPTS --host $HOST production 2>&1 | logger --tag docker-worker
+/usr/local/bin/start-worker /etc/start-worker.yml 2>&1 | logger --tag docker-worker
--- a/config.yml
+++ b/config.yml
@@ -61,6 +61,9 @@ defaults:
 
   alivenessCheckInterval: 30000
 
+  # by default, run one task at a time
+  capacity: 1
+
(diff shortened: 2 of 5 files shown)


djmitche:
This uses the existing `host` mechanism, but treats taskcluster-worker-runner as a "host".  This is the basic idea of the tool: the worker implementation shouldn't need to have special-case code for each cloud it might be run in.

The defaulting of `capacity` is because at the moment worker-manager doesn't automatically specify capacity to the worker.  I'd like to leave that as just plain old worker configuration rather than anything w-m knows how to specify.

I don't want to deploy this to production yet, since it misses some of the features that docker-worker already has, like termination notification, but I'd like to get it landed all the same so that I can continue to work on this functionality.

deploy/template/usr/local/bin/start-docker-worker
```diff
@@ -1,6 +1,3 @@
 #!/bin/bash
 
-DOCKER_WORKER='node /home/ubuntu/docker_worker/src/bin/worker.js'
-DOCKER_WORKER_OPTS=
-
-$DOCKER_WORKER $DOCKER_WORKER_OPTS --host $HOST production 2>&1 | logger --tag docker-worker
+/usr/local/bin/start-worker /etc/start-worker.yml 2>&1 | logger --tag docker-worker
```
walac: Instead of changing the existing script, could you please create another script? Since it is not production ready yet, it may impact deployments in the mean time.

djmitche: Oh, yes, that's silly of me!  Maybe I'll keep this alive on a branch for the moment -- the alternative would be to update the packer build process to output another two sets of AMIs.  I'll also focus on getting feature parity here.

djmitche:
I'll open a new PR under [bug 1558530](https://bugzilla.mozilla.org/show_bug.cgi?id=1558530) when I get to feature-parity.

PR #472, opened 2019-07-17
Implement experimental support for Google Computer Engine

walac:
[Bug 1567166](https://bugzilla.mozilla.org/show_bug.cgi?id=1567166)

PR #465, opened 2019-06-27
Bug 1558530, 1546686 - run under tc-worker-runner

djmitche:
This includes #464. 

This is about all of the "protocol" we'll ever support in docker-worker, but does bring the tc-worker-runner "host" up to the same functionality level as the other hosts (specifically, supporting spot-instance shutdown).

PR #473, opened 2019-07-18
Implement support for building GCP images

djmitche:
My changes on top of @walac's implementation -- a followon from #472.

[Bug 1567166](https://bugzilla.mozilla.org/show_bug.cgi?id=1567166)


walac:
Does this require #465 merged?

deploy.sh
```diff
@@ -2,14 +2,19 @@
 
 export BUILD_TARGET=$1
 if [ -z "$BUILD_TARGET" ]; then
-    echo "USAGE: ./deploy.sh <target>   (base, app, etc.)" >&2
+    echo "USAGE: ./deploy.sh <target> [options]  (base, app, etc.)" >&2
+    echo " --test -- (for legacy deployment only) only deploy to ami-test, not all workers"
+    ehco " --only='..' -- only build these packer builders"
```
walac: +1

walac: s/`ehco`/`echo`
Also would be nice if the help message provided the options, and how should we pass them (space or comma separated?)

djmitche:
> Does this require #465 merged?

Nope.  *I* would like that merged, but the two PRs are independent :)

deploy.sh
```diff
@@ -51,6 +68,23 @@ if ! $found; then
 fi
 echo "Using AWS account $AWS_ACCOUNT" >&2
 
+export GCP_PROJECT_ID
+case "$DEPLOYMENT" in
+    #taskcluster-net) GCP_PROJECT_ID=linux64-builds;;
```
walac: I assume this is commented intentionally.

deploy.sh
```diff
@@ -51,6 +68,23 @@ if ! $found; then
...
+    dustin-dev) GCP_PROJECT_ID=dustin-dev-3;;
+    *)
+        echo "No GCP project defined for this environment" >&2
+        exit 1;;
+esac
+
+gcloud projects describe $GCP_PROJECT_ID >/dev/null
+if [ $? != 0 ]; then
```
walac: nit: we can shorten this with:
```sh
if !gcloud projects describe $GCP_PROJECT_ID >/dev/null; then
```

deploy/bin/build
```diff
@@ -123,6 +138,7 @@ esac
 if [ -f packer-artifacts.json ]; then
   jq '
     .builds[]
+    | select(.builder_type == "amazon-ebs")
```
walac: I don't know what I had in mind what I thought using `jq` would be a good idea.

deploy.sh
djmitche: Good point.  I was thinking of actually switching this to something like `--cloud=aws` to only build specific clouds, and then translating that to packer builders later.

deploy.sh
djmitche: It was, but probably unnecessarily.

deploy/bin/build
djmitche: It works fine, it just doesn't fit in my brain :)

djmitche:
Hah, now I can't push to your branch..

PR #474, opened 2019-07-19
Implement support for building GCP images

djmitche:
My changes on top of @walac's implementation -- a followon from #472, #473 with a different head branch.

[Bug 1567166](https://bugzilla.mozilla.org/show_bug.cgi?id=1567166)


djmitche:
(this just has the one additional commit - none of the others were changed)

PR #480, opened 2019-07-29
Bug 1527107: Fix CVE-2019-5736

walac:
docker-ce version 18.06.3 fixes this vulnerability.


walac:
@dustin from [this blog post](https://blog.docker.com/2019/02/docker-security-update-cve-2018-5736-and-container-security-best-practices/) I can confirm that version `18.06.3` fixes the vulnerability.

deploy/packer/app/scripts/deploy.sh
```diff
@@ -108,6 +108,6 @@ User=root
 RequiredBy=graphical.target
 EOF'
 
-sudo systemctl enable docker-worker.service
-sudo systemctl enable side-containers.service
-sudo systemctl daemon-reload
+#sudo systemctl enable docker-worker.service
+#sudo systemctl enable side-containers.service
+#sudo systemctl daemon-reload
```
djmitche: intentional?

deploy/packer/base/scripts/packages.sh
```diff
@@ -2,9 +2,9 @@
 
 set -e -v
 
-DOCKER_VERSION=5:18.09.5~3-0~ubuntu-bionic
-KERNEL_VER=4.15.0-47-generic
-V4L2LOOPBACK_VERSION=0.12.0
+DOCKER_VERSION=18.06.3~ce~3-0~ubuntu
+KERNEL_VER=4.4.0-1014-aws
```
djmitche: will this work in GCP and Packet?

src/lib/host/aws.js
```diff
@@ -123,6 +121,8 @@ module.exports = {
       spawn('shutdown', ['-h', 'now']);
     }
 
+    log('read userdata', { text: _.omit(userdata, ['securityToken']) });
+
```
djmitche: In theory this _host_ isn't used anymore.. is it?

deploy/packer/base/scripts/packages.sh
djmitche: I guess the question behind that would be: has this built successfully for those clouds as well?

walac: packet has a different build process, for it we will stick to Ubuntu 18.04. I didn't test it on GCP, but for EC2 it works.

src/lib/host/aws.js
walac: What do you mean?

djmitche: We use the `worker-runner` host for everything, so the aws host shouldn't matter (in fact, I probably should have deleted it).

deploy/packer/base/scripts/packages.sh
djmitche: Please test on GCP :)

src/lib/host/aws.js
walac: Ah, this is just part of the `git revert` command.

djmitche: Ah, OK.

djmitche:
I realized last night that this is a revert of the *upgrade* to 18.04 that was never deployed.  So my concerns about the downgrade are a little less critical and don't block landing.

The GCP support does block landing, though.

I'm still worried that we can't run a supported OS, though.  We may not be so lucky with the next CVE...

walac:
This is a short term solution, sooner or later we will have to fix the issues coming with OS upgrade.

djmitche:
OK, that's a relief :)

deploy/packer/base/scripts/packages.sh
walac: Ok, update the patch to build GCP successfully

PR #488, opened 2019-07-31
Bug 1569856: Provide a default disk limit for volume cache

walac:
If the work types has disk limit configuration for volume caches, it may
grows up to let the system out of space.

### DIFF
--- a/src/lib/host/taskcluster-worker-runner.js
+++ b/src/lib/host/taskcluster-worker-runner.js
@@ -37,6 +37,10 @@ module.exports = {
       throw new Error('No config file found');
     }
     const content = fs.readFileSync(configFile, 'utf8');
-    return JSON.parse(content);
+    return Object.assign({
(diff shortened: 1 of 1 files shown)


src/lib/host/taskcluster-worker-runner.js
```diff
@@ -37,6 +37,10 @@ module.exports = {
       throw new Error('No config file found');
     }
     const content = fs.readFileSync(configFile, 'utf8');
-    return JSON.parse(content);
+    return Object.assign({
+      capacityManagement: {
+        diskspaceThreshold: 30000000000,
```
djmitche: The default for this is already set (but to a different value) at
https://github.com/taskcluster/docker-worker/blob/4cad67ba1759c5676a5c52ad115725a7fb6f5067/config.yml#L67-L68
Wouldn't it make more sense to update that value, rather than default it here?

PR #495, opened 2019-08-16
Shutdown instance on worker exit

walac:
The worker should run for the entire instance life. If it exits sooner,
it means something has gone wrong. In this case, we turn off the
instance and wait for the provisioner to kill it.


walac:
> This looks good, but I'm confused about your comment: spot instances are "terminated" when they halt, so there's nothing for the provisioner to do here.

They provisioner terminates them, I believe.

djmitche:
John was pretty adamant that the provisioner doesn't do so.  Anyway, I think the effect is fine, so land it!

PR #501, opened 2019-09-02
Bug 1405711: Don't pick a new task after spot termination event r=dustin

walac:
When we receive a spot termination notice from the worker-runner, the
shutdown manager set the configured capacity to zero and the current
running tasks are killed with `worker-shutdown` state.

`TaskListener` caches the `capacity` configuration, so when the
shutdown manager updates it, `TaskListener` continues reading a stale
value. It then picks another task to run, and when the spot is about
to terminate, the Linux kernel sends a SIGTERM to the running container
and makes the new running task fails.

We fix it by not caching the capacity configuration anymore,
`TaskListener` now reads it directly from the configuration object.

### DIFF
--- a/src/lib/task_listener.js
+++ b/src/lib/task_listener.js
@@ -19,7 +19,6 @@ class TaskListener extends EventEmitter {
   constructor(runtime) {
     super();
     this.runtime = runtime;
-    this.capacity = runtime.capacity;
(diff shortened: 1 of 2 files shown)


walac:
The task failure it unrelated, it worked when I reran it.

PR #502, opened 2019-09-03
Bug 1302596: Retry image download when it fails

walac:
We already have retries for when the download fails, but not when the
downloaded image is corrupted.

We add a retry for the general `ensureImage()` function, for edge cases
like above.

As we are not retrying because of possible network temporary outage, we
configure the waiting time to be a small value.


walac:
The failed test is unrelated and you will be addressed in a follow-up PR.

PR #505, opened 2019-09-10
Remove encrypted variables support

walac:
The version we use of openpgp has security vulnerabilities, and it is
used only for the deprecated (and not used anymore) envrironment
variables feature.

So, we remove the encrypted variables support and the openpgp package.


walac:
Good catch on the payload schema. If there is a task out there using this feature, it must be well hidden, I never saw a task using encrypted env. We can announce on tc mail list.

schemas/v1/payload.json
```diff
@@ -160,15 +160,6 @@
...
     },
-    "encryptedEnv": {
-      "title": "List of encrypted environment variable mappings.",
-      "description": "List of base64 encoded asymmetric encrypted environment variables. See /docs/reference/workers/docker-worker/environment#encrypted-environment-variables",
-      "type": "array",
-      "items": {
-        "title": "Base64 encoded encrypted environment variable object.",
-        "type": "string"
-      }
-    },
```
djmitche: I forgot we don't have `additionalProperties: false` here, so this actually *won't* generate an error if the parameter is still used.  Oh well :(

walac: That comes from our old try on adding this restriction and breaking a bunch of ill-formed task payloads.

djmitche: Yup, I understand :)

At least with this change the variable won't be in the docs anymore.

PR #520, opened 2019-10-28
Run packet.net instance with taskcluster-worker-runner

walac:
As worker-manager packet provider isn't production ready yet, we run
packet.net image instances through standalone provider.

We have a new `tc-w-runner-config` systemd service that creates the
`/etc/start-worker.yml` file.

Notice the image doesn't ship with the SSL certificate, which means that
if the option `secureLiveLogging` is true, the worker won't pick any
task to execute.

Finally, the `ubuntu_18_04-base` image is now stored in the
`docker.io/taskcluser` docker namespace, avoiding the need for building
the base image before creating a new docker packet image.

### DIFF
--- a/Dockerfile.packet
+++ b/Dockerfile.packet
@@ -1,4 +1,4 @@
-FROM ubuntu_18_04-base
+FROM taskcluster/ubuntu_18_04-base
 MAINTAINER Wander Lairson Costa <wcosta@mozilla.com>
 LABEL Description="docker-worker packet.net image" Vendor="Mozilla"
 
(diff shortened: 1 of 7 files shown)


djmitche:
> As worker-manager isn't production ready yet, we run packet.net image instances through standalone provider.

What does this mean?  W-M is in production now :)

walac:
> What does this mean? W-M is in production now :)

Not with packet provider, but I will update the comment

djmitche:
Ah, indeed, thanks.

PR #522, opened 2019-10-30
Bug 1590226: Add worker activity logging

walac:


src/lib/runtime.js
```diff
@@ -69,7 +69,25 @@ Runtime.prototype = {
...
+    this.log('WORKER_METRICS', {
+      eventType,
+      worker: 'docker-worker',
+      workerPoolId: this.workerType,
+      workerId: this.workerId,
+      timestamp,
+      region: this.region,
+      instanceType: this.instanceType,
+      taskId: task.status.taskId,
+      runId: task.status.runId,
```
djmitche: These will fail when `task` is `{}`, right?

walac: Yes, thanks. I fixed locally!

djmitche: I came, I saw, i r+'d :)

PR #530, opened 2019-11-27
Bug 1599564 - only purge caches when claiming tasks

djmitche:
Previous behavior was to purge caches on every claimWork call, so about
once every 25 seconds for an idle worker.

The new behavior matches generic-worker.


walac:
@djmitche purge tests are failing

djmitche:
:eyes:

djmitche:
all green :D

PR #532, opened 2019-12-11
Add a capability disableSeccomp to disable the seccomp system call filter.

jschwartzentruber:
This is required to run `rr` inside a container, as described here: https://github.com/mozilla/rr/wiki/Docker

I haven't been able to run the test locally, mocha says: "pending". I'm hoping it'll be run in CI.


config.yml
```diff
@@ -33,6 +33,7 @@ defaults:
     relengAPIProxy:
       image: 'taskcluster/relengapi-proxy:2.3.1'
       token: !env RELENG_API_TOKEN
+    disableSeccomp: false
```
walac: I think I was not clear in my review. What I meant was a configuration that allows the worker to process `disableSeccomp: true` feature. Look at the `allowPrivileged` configuration for an example on how to do so.

jschwartzentruber: Sorry, I still don't understand. The example config you linked to is for `allowPrivileged: false`, so where should I look for an example config where `allowPrivileged` is true?

I modeled the change after `allowPtrace` which is a closer analogue, and I don't see that referenced in config.yml.

walac: The role of `allowPrivileged` is adding another layer of security to the worker-pool. It is a worker-pool configuration that when it is `true`, [it doesn't allow task to run as a privileged container, even if it requests in the payload](https://github.com/taskcluster/docker-worker/blob/master/src/lib/task.js#L327-L329). Maybe it is easier if you implement `disableSeccomp` as a capability instead of a feature.

jschwartzentruber: Thanks, I understand now. I'll take a look.

walac:
It took a while but the tasks now run, it feels like disableSeccomp tasks are failing

walac:
@djmitche another one:
```
diff --git a/test/dockerworker.js b/test/dockerworker.js
index 8ed1eef8..7ef763d6 100644
--- a/test/dockerworker.js
+++ b/test/dockerworker.js
@@ -64,7 +64,7 @@ class DockerWorker {
           global.asyncDump ? '--require /worker/src/lib/async-dump' : '',
           '/worker/src/bin/worker.js',
           '--host test',
-          '--worker-group', 'random-local-worker',
+          '--worker-group', 'docker-worker',
           '--worker-id', this.workerId,
           '--provisioner-id', this.provisionerId,
           '--worker-type', this.workerType,
```

djmitche:
Not my PR, but I pushed that change :)

walac:
> Not my PR, but I pushed that change :)

Ops, I commented in the wrong PR, it actually applies to yours, sorry

PR #533, opened 2019-12-11
Add a device to mount the host /dev/shm inside the container.

jschwartzentruber:
We do this when fuzzing, since Firefox needs a lot of space in /dev/shm and sizing it properly is not straightforward.

### DIFF
--- a/src/lib/devices/shared_memory_device_manager.js
+++ b/src/lib/devices/shared_memory_device_manager.js
@@ -1,5 +1,4 @@
 const Debug = require('debug');
-const fs = require('fs');
 
 let debug = Debug('taskcluster-docker-worker:devices:sharedMemoryManager');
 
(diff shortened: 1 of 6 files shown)


src/lib/devices/shared_memory_device_manager.js
```diff
@@ -0,0 +1,48 @@
...
+    return device;
+  }
+
+  getAvailableDevices() {
+    return this.devices;
+  }
+}
+
+class SharedMemoryDevice {
+  constructor(path) {
```
walac: Why does the constructor accept a `path` parameter if only `/dev/shm` is allowed?

walac:
I create a [PR](https://github.com/mozilla/community-tc-config/pull/174) in the community repo to add the shared memory capability scope.

walac:
There is something wrong with the linter task, but I am quite sure it is unrelated to your patch. I am going to investigate it.

walac:
Ok, I found out what's wrong. As you added the `hostSharedMemory` to `.taskcluster.yml` this created a chicken-egg problem, since the currently deployed docker-worker obvisiously doesn't support this device. I wonder how tests were passing before.

jschwartzentruber:
How do we get around this chicken-egg problem? Is there anything for me to do?

PR #538, opened 2020-02-21
Fix instanceBoot time

catlee:
os.uptime() returns a number of seconds since boot, so we need to multiply this by 1000 to get a proper offset from Date.now().


walac:
Thanks!

PR #540, opened 2020-02-25
Bug 1520824: add kvm device

walac:
When enabled, it exposes /dev/kvm to the container, allowing to run
emulators with hardware virtualization support.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -9,6 +9,8 @@ defaults:
       enabled: true
     hostSharedMemory:
       enabled: true
+    kvm:
+      enabled: true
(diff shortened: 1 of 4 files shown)


walac:
> Do we want to be able to control access to this feature on a per-worker-pool basis? I think we do that for privileged?

`/dev/kvm` doesn't have security implications according to my research

> It looks like this would allow concurrent tasks to both talk to /dev/kvm - is that true, and intended?

It causes no harm it being shared, afaik. Furthermore, it is implemented for one specific use case.

PR #541, opened 2020-02-27
Delegate instance shutdown to the host configuration

walac:
Now each host configuration must implement a "shutdown" function. In the
case of worker-runner, it will call worker-manager removeWorker API and,
if it fails, falls back to a "system shutdown".

This must fix the problem in which instances start to shutdown but
stalls in the middle of the process, hanging forever.

Depends on: https://github.com/taskcluster/taskcluster/pull/2425

### DIFF
--- a/src/lib/host/packet.js
+++ b/src/lib/host/packet.js
@@ -112,4 +112,7 @@ module.exports = {
   billingCycleUptime() {
     return os.uptime();
   },
+  async shutdown() {
+    // do nothing
+  },
 };
--- a/src/lib/host/taskcluster-worker-runner.js
+++ b/src/lib/host/taskcluster-worker-runner.js
@@ -15,6 +15,7 @@ module.exports = {
     // docker-worker doesn't support a finish-your-tasks-first termination,
     // so we ignore that portion of the message
     protocol.addCapability('graceful-termination');
+    protocol.addCapability('shutdown');
     protocol.on('graceful-termination-msg', () => {
       gracefulTermination = true;
     });
@@ -39,5 +40,12 @@ module.exports = {
     }
     const content = fs.readFileSync(configFile, 'utf8');
     return JSON.parse(content);
-  }
+  },
+
+  async shutdown() {
+    if (!await protocol.capable('shutdown')) {
+      throw new Error('Shutdown called but worker-runner doesn\'t support this capability');
+    }
+    protocol.send({type: 'shutdown'});
+  },
 };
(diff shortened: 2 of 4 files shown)


src/lib/worker-runner-protocol.js
```diff
@@ -85,6 +85,10 @@ class Protocol extends EventEmitter {
     this.transport.send(message);
   }
 
+  shutdown() {
+    this.send({type: 'shutdown', capabilities: [...this.capabilities]});
+  }
```
djmitche: This message type will require adding a  new capability to `supportedCapabilities`.

I'd prefer to have the protocol only implement `send`, and have the Host implementation call that.  Also, it should consult the protocol's capabilities and only send the message if it's supported.  I'm not sure what to do if it's not supported, though. Do we take that as an indication that this is not a system that should be shut down (static/standalone), in which case maybe `process.exit(0)` is the right choice?  Or that worker-runner doesn't support this functionality and docker-worker should do it directly, in which case invoking `shutdown -h now` is the right choice?

walac: > This message type will require adding a new capability to supportedCapabilities.

Ok

> I'd prefer to have the protocol only implement send, and have the Host implementation call that.

Makes sense

> Also, it should consult the protocol's capabilities and only send the message if it's supported. I'm not sure what to do if it's not supported, though

Do you mean sending a message to `worker-runner` querying the capabilities? I am thinking about when `worker-runner` receives a message type it doesn't support, it could reply with a message type like `unsupported-<message-type-received>`

djmitche: > Do you mean sending a message to worker-runner querying the capabilities? I am thinking about when worker-runner receives a message type it doesn't support, it could reply with a message type like unsupported-<message-type-received>

The protocol already includes a capability-negotiation phase, so you can just query `proto.capable('shutdown')`.

walac:
Yeah, but iiuc, the negotiation phase happens before we add the "shutdown"
capability, doesn't it? Well, I could move the Capability.Add call to the
SetProtocol method.

Em sex, 28 de fev de 2020 16:23, Dustin J. Mitchell <
notifications@github.com> escreveu:

djmitche:
I looked at that and I don't *think* it's the case, but I could be wrong.  `runner/run.go` calls `StartWorker` before it starts the protocol, and the `Add` calls as you've written them happen between those two things.  At least, that's what I remember from looking yesterday..

PR #542, opened 2020-02-28
Wait until the worker-runner protocol has initialized

djmitche:
This causes `capable` checks to block until the protocol is established.
It also makes the `capable` method async, but at this point it has no
uses outside of tests.

A downside of this approach is that anything calling `capable` will hang forever if docker-worker is run without worker-runner.  In generic-worker it was important to maintain the ability to run the worker "alone", so we added a `--with-worker-runner` command-line flag which indicates to generic-worker that it's running under worker-runner.  When that flag is missing, generic-worker just assumes all capabilities aren't set.  We could do something similar for docker-worker if that's useful.  What do you think?

### DIFF
--- a/src/lib/worker-runner-protocol.js
+++ b/src/lib/worker-runner-protocol.js
@@ -75,7 +75,13 @@ class Protocol extends EventEmitter {
       this.emit(event, msg);
     });
 
-    this.on('welcome-msg', msg => this._handleWelcome(msg));
+    // create a promise that will resolve when we have received a welcome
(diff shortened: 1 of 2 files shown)


walac:
>  What do you think?

docker-worker already has a command-line argument to [set the host](https://github.com/taskcluster/docker-worker/blob/master/src/bin/worker.js#L72)

djmitche:
Oh -- good point!  So in the absence of `--host taskcluster-worker-runner` it wont' block.

PR #544, opened 2020-03-11
exit if cloud is unknown

djmitche:


djmitche:
..or, if this directory isn't used anymore and we're using monopacker, should I make a PR to delete it?

djmitche:
This directory is still used for firefox-ci.  It's a tiny nit though :)

PR #547, opened 2020-03-14
Update mkdirp

djmitche:
This gets partway to upgrading minimist, but several things still depend
on an old version of minimist.  Happily none of those are used in a
context where they parse untrusted command lines.


walac:
It feels like the new version of `mkdirp` isn't backward compatible.

djmitche:
Oh, that was silly of me.  I'll have a look.

djmitche:
Let's see how that does..

PR #550, opened 2020-03-30
Bug 1625317: Handle exception when logging error to monitor

walac:


src/bin/worker.js
```diff
@@ -122,6 +119,16 @@ program.parse(process.argv);
     config = _.defaultsDeep(targetConfig, config);
   }
 
+  process.on('unhandledRejection', async (reason, p) => {
+    console.error(`Unhandled rejection at ${p}.\n${reason.stack || reason}`);
+    if (host && host.shutdown) {
+      await host.shutdown();
+    } else {
+      spawn('shutdown', ['-h', 'now']);
```
petemoore: This might be helpful if an interactive task is running (so that the interactive session is alerted to the reason for the shutdown):

```suggestion
      spawn('shutdown', ['-h', 'now', `docker-worker shutdown due to unhandled rejection ${p}`]);
```

src/bin/worker.js
```diff
@@ -122,6 +119,16 @@ program.parse(process.argv);
     config = _.defaultsDeep(targetConfig, config);
   }
 
+  process.on('unhandledRejection', async (reason, p) => {
```
djmitche, line 122: Would it make sense to use `process.once` here, so if there's a second unhandled error it just kills the process?

PR #553, opened 2020-04-20
Bug 1631049: Handle failure to update credentials in the proxy

walac:
If the request to update the taskcluster-proxy credentials fails, the
reject promise is unhandled and the nodejs process exits.

To mitigate this issue, we:

1) Use exponential backoff retry when updating the credentials
2) Handles the promise rejection in case all retries fail

Notice we keep the task running because this only impacts the task if it
makes a request to http://taskcluster/ URL.


walac:
> it's because the proxy's been killed or otherwise failed?

I believe it is random failure.

PR #555, opened 2020-04-22
Remove deployment scripts

walac:
Deployment scripts has been move to their own repo [1]

[1] https://github.com/taskcluster/docker-worker-deploy


walac:
The failure is a known intermittent.

PR #556, opened 2020-04-23
Cleanup unused packages

walac:
Part of the preparation for monorepo migration.


walac:
Due the github issues the tests failed at first, but seems to be fine now.

PR #557, opened 2020-04-30
Upgrade vagrant files

walac:
With deployment scripts in its own repo, we need to upgrade vagrant
configuration to look for them in the right place.


milescrabill (approved):
This works but would be cleaner if we set the shell path. Right now we’re invoking bash within bash on each line. Could be worth explaining too.

